PR #9, opened 2020-12-31
Move example webpack config into examples/, add README

pastelmind:
This PR moves the `webpack.config.js` into `examples/`, and adds `examples/README.md`.

## Why

The current webpack configuration is used to build the example code, not libram itself.

- If we're not going to use webpack to build libram, this file is going to confuse people simply by existing.
- If we're going to use webpack to build libram, then this file needs to be moved somewhere else, and we should put a proper build config in its place.

Either way, the example build config doesn't belong in the project root.


gausie (approved):
Fantastic, thank you

PR #10, opened 2020-12-31
Add support for downstream JS users/bundlers

pastelmind:
This PR adds support for libram users who use JavaScript but not TypeScript:

- Expose JS files via `pkg.module` so that JS bundlers like Webpack and Rollup can pick them up
- Remove `pkg.main`, since libram cannot be executed in Node.js (yet)

There are other quality-of-life improvements:

- Fix `yarn run build` (specify `include` in `tsconfig.json`) (982deff55e2dcb411f3e2e759e85a0ca7cd77147)
- Add JS/D.TS files to `.gitignore`
- Generate D.TS files, so that people can browse the type definitions without having to also look at implementation code
- Expose D.TS files through `pkg.types`

@gausie If this PR makes it in, we should make sure to run `yarn run build` before `yarn publish`.


gausie:
I usually like to keep the package.json agnostic to whether the developer is using `npm` or `yarn` and so tend to use `npm run` (because everyone has `npm` installed) in any `scripts`. That doesn't work here for whatever reason which is strange so I've left as is. Great job!

gausie (approved):

PR #11, opened 2020-12-31
Add yarn clean/lint script

pastelmind:
- Add `yarn run clean` which deletes all generated JS/D.TS files.
- Add `yarn run lint` which runs ESLint on TypeScript code. (desired for #13)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,4 +8,4 @@ jobs:
       - name: Install modules
         run: yarn
       - name: Run ESLint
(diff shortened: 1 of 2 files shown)


gausie:
We don't need tsx ever

pastelmind:
Sure.

I also remembered to make the workflow to use `yarn run lint` instead of running `yarn run eslint` directly.

gausie:
Thank you! That's fantastic.

gausie (approved):

Issue #13, opened 2020-12-31
Add a CONTRIBUTING.md


pastelmind:
Things to add:

- [x] Setting up dev environment
   - [x] Yarn version?
- [ ] Recommended editor (VS Code?)
- [ ] Preferred code style (*might be unnecessary?*)
- [ ] How to run linting
   - [ ] We need to set up a `yarn run lint` script first (#11)
- [ ] How to run tests locally

PR #14, opened 2020-12-31
docs: Add CONTRIBUTING.md

pastelmind:
Adds bare-bones Contribution Guidelines. We can expand it later as we see fit.

This does 90% of #13, except the "how to run tests" part, because we don't have tests yet.


gausie (approved):
Thank you!

PR #15, opened 2021-01-02
Add combat stuff.

phulin:


pastelmind:
It might be time to discuss development tooling.

@gausie and I have been using Yarn. @phulin I assume you are using NPM, since you added `package-lock.json`.

Maintaining both `package-lock.json` and `yarn.lock` would be tedious and error-prone. I suggest we agree to adopt one package manager.

(I mentioned it in #13 but it's awaiting a review.)

phulin:
Oh, this is probably just me having no idea what I'm doing with JS tooling. I'll switch to Yarn.

pastelmind (dismissed):
It compiles fine for me. Good work!

gausie:
This is amazing. Minor points

* Could you add some jsdocs to the functions you've written? They are super cool but I think they need to be well documented to be well used.
* Did you run prettier on this? `yarn run format` should do it.
* Can you remove `print`s? I think we should consider adding debug logging in the future but I think having unavoidable prints is bad form.

phulin:
Yep, yarn run format only touches files not in my PR. Should probably address that in a separate PR.

phulin:
Adding a couple more things and then it should be ready to merge in my view.

gausie (approved):
Fantastic. I don't think we need types in the jsdoc but this is great!

PR #16, opened 2021-01-03
Exclude source .ts files from being published

pastelmind:
This is a follow-up to #10.

When I set up `yarn run build` to generate `.js` and `.d.ts` files, I forgot about updating `pkg.files`. Now, both the build artifacts (`.js` and `.d.ts`) and the TypeScript source files (`.ts`) are published.

Once libram is installed as a dependency, the TypeScript files don't do anything but take up disk space. I'm submitting a patch that excludes them from the publishing process.

Tested with `npm pack`, `npm publish --dry-run` and `yarn pack` to ensure that only the build artifacts are included when publishing. I also installed the generated tarball in a sample project to verify that the type definitions and code still work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,10 @@
     "prepare": "patch-package",
     "prepublishOnly": "yarn run build"
   },
-  "files": ["src/"],
+  "files": [
+    "src/**/*.js",
+    "src/**/*.d.ts"
+  ],
   "devDependencies": {
     "@babel/cli": "^7.12.1",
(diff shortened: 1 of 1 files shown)


pastelmind:
Also, just for fun, here's an image from the sample project:
image

gausie:
> Also, just for fun, here's an image from the sample project:
> 
> image

I'd really like to not include he.js actually, it seems non-relevant for libram. Sadly the node html parser insists on it.

gausie (approved):

PR #17, opened 2021-01-06
Fix circular imports

pastelmind:
Fix circular imports. I used [madge](https://github.com/pahen/madge) to identify circular imports and manually removed them.

- Most issues are solved by simply importing from the module that defines the function.
- `SpookyPutty.ts` and `RainDoh.ts` had some functions that were referencing each other, so extract them to `putty-likes.ts` which serves as a coalescing module of sorts.

Incidentally, this did not solve the bundling issue where tree shaking fails and the entirety of `libram` and `node-html-parser` is included in the bundle. I suspect it has something to do with `node-html-parser` itself.

Edit: See #18 for enabling tree shaking.

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -26,8 +26,6 @@ import {
 
 import { $class } from "./template-string";
 import { get } from "./property";
-import { getRainDohBlackBoxCopiesMade } from "./resources/RainDoh";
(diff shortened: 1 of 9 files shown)


gausie (approved):
I think `putty-likes` may end up with a better home in the future, but this needs to be merged. Thank you!

PR #19, opened 2021-01-17
Add more free fight capabilities.

phulin:
This adds easy code to use LOV, Witchess, and Chateau painting free fight sources.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,5 @@
+node_modules/
+dist/
+.vscode/
+src/**/*.js
+src/**/*.d.ts
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,4 @@
 node_modules/
 dist/
 .vscode/
-src/**/*.js
-src/**/*.d.ts
+.DS_Store
--- a/src/resources/ChateauMantegna.ts
+++ b/src/resources/ChateauMantegna.ts
@@ -1,0 +1,19 @@
+import { runCombat, visitUrl } from "kolmafia";
+import { get } from "../property";
+
+export function have(): boolean {
+  return get("chateauAvailable");
+}
+
+export function paintingMonster(): Monster | null {
+  return get("chateauMonster") ?? null;
+}
+
+export function paintingFought(): boolean {
+  return get("_chateauMonsterFought");
+}
+
+export function fightPainting(): string {
+  visitUrl("place.php?whichplace=chateau&action=chateau_painting", false);
+  return runCombat();
+}
--- a/src/resources/ChateauMantegna.ts
+++ b/src/resources/ChateauMantegna.ts
@@ -6,7 +6,7 @@ export function have(): boolean {
 }
 
 export function paintingMonster(): Monster | null {
-  return get("chateauMonster") ?? null;
+  return get("chateauMonster");
 }
 
 export function paintingFought(): boolean {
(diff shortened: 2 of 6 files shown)


pastelmind:
Sorry, but I'm unfamiliar with LOV/Chateau/whatever iotm mechanics that were introduced while I wasn't around. Maybe @gausie is better qualified for reviewing this.

.eslintignore
```diff
@@ -0,0 +1,5 @@
+node_modules/
+dist/
+.vscode/
+src/**/*.js
```
gausie, line 4: Do we need this anymore now that we've changed the outdir?

src/resources/ChateauMantegna.ts
```diff
@@ -0,0 +1,19 @@
+import { runCombat, visitUrl } from "kolmafia";
+import { get } from "../property";
+
+export function have(): boolean {
+  return get("chateauAvailable");
+}
+
+export function paintingMonster(): Monster | null {
+  return get("chateauMonster") ?? null;
```
gausie, line 9: Does this not return `null` by default? I tried to make the `get` code do that for monster keys

src/resources/TunnelOfLove.ts
```diff
@@ -41,3 +38,76 @@ export const LovEnamorang = new Copier(
...
+type LOVEquipment = "LOV Eardigan" | "LOV Epaulettes" | "LOV Earring";
+type LOVEffect = "Lovebotamy" | "Open Heart Surgery" | "Wandering Eye Surgery";
+type LOVExtra =
+  | "LOV Enamorang"
+  | "LOV Emotionizer"
+  | "LOV Extraterrestrial Chocolate"
+  | "LOV Echinacea Bouquet"
+  | "LOV Elephant"
+  | "toast"
+  | null;
```
gausie, line 51: Why null here but not the other options? In case you're going for the trophy or something

src/resources/Witchess.ts
```diff
@@ -1,8 +1,36 @@
+import { myHash, runChoice, runCombat, toInt, visitUrl } from "kolmafia";
 import { haveInCampground } from "../lib";
+import { get } from "../property";
 import { $item } from "../template-string";
 
-export const item = $item`Witchess Set`;
```
gausie, line 4: I kinda liked being able to do `witchess.item`, but I can see why we might want to remove this

src/resources/Witchess.ts
```diff
@@ -1,8 +1,36 @@
...
-  return haveInCampground(item);
+  return haveInCampground($item`Witchess Set`);
+}
+
+export function fightsDone(): number {
+  return get("_witchessFights");
+}
+
+export function fightPiece(piece: Monster): string {
+  if (
```
gausie, line 15: Would be cool to add a check that its a valid witchess monster

gausie (approved):
This is mergeable though I had some small points.

Once this is merged I'll bump and publish!

src/resources/TunnelOfLove.ts
phulin: These are all the options, I think (null means take nothing). Trophy support will have to come later.

src/resources/Witchess.ts
phulin: fixed

src/resources/Witchess.ts
phulin: done

PR #20, opened 2021-01-21
Build to dist directory, not src.

phulin:
This makes importing scripts prioritize importing the compiled JS over the source TS, which in turn means that those scripts don't need to have all the babel and typescript plugins we have turned on (e.g. decorators).

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,23 +2,23 @@
   "name": "libram",
   "version": "0.0.4",
   "description": "JavaScript helper library for KoLmafia",
-  "module": "src/index.js",
-  "types": "src/index.d.ts",
+  "module": "dist/index.js",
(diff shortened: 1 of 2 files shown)


pastelmind:
Approved.

That said, are we going to allow downstream users to "cherry pick" imports by importing specific files? I consider our directory/file structure an implementation detail, and users should not try to dive into the directories. All imports should be made with `import {X} from "libram"`. We have tree shaking enabled (see #18), so importing from the package root won't impact bundle size.

pastelmind (approved):

phulin:
I agree - I don't think we want to allow imports from subdirectories. If we want to enforce that we'll have to bundle.

gausie:
We can enforce this using the `exports` key in package.json too

PR #22, opened 2021-01-27
Reorganize resources and add SongBoom.

phulin:
Reorganize resources into year-based folders. I expect these to fill up as we continue adding things.

Oh, and I turned sort-imports on in eslint. Had been annoying me. Can discuss if we don't want that.


gausie (approved):
Perfect, thank you. And I love sort-imports, thanks for that too

PR #23, opened 2021-01-27
Add basic mood support.

phulin:
Add support for defining moods in TS. 

This is pretty basic for now - I'll add more planning stuff later. I want to add a pre-casting framework for speed reasons.

### DIFF
--- a/src/mood.ts
+++ b/src/mood.ts
@@ -186,13 +186,14 @@ export class Mood {
   /**
    * Add an effect to the mood, with casting based on {effect.default}.
    * @param effect Effect to add.
+   * @param gainEffect How to gain the effect. Only runs if we don't have the effect.
    */
-  effect(effect: Effect): void {
+  effect(effect: Effect, gainEffect?: () => void): void {
     const skill = toSkill(effect);
-    if (skill !== $skill`none`) {
+    if (!gainEffect && skill !== $skill`none`) {
       this.skill(skill);
     } else {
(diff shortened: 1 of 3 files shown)


src/mood.ts
```diff
@@ -0,0 +1,227 @@
...
+    return result;
+  }
+
+  elements: MoodElement[] = [];
+
+  /**
+   * Add a skill to the mood.
+   * @param skill Skill to add.
+   */
+  skill(skill: Skill): void {
```
gausie, line 182: You could make all these chainable by returning `this` instead of `void`!!

src/mood.ts
```diff
@@ -0,0 +1,227 @@
...
+    return haveEffect(this.effect) > ensureTurns;
+  }
+}
+
+/**
+ * Class representing a mood object. Set options using the static methods, and add mood elements using the instance methods.
+ */
+export class Mood {
+  static songSlots: Effect[][] = [];
+  static useSausages = true;
```
gausie, line 156: This should probably be an optional argument to the Mood constructor. And done via that in general rather than modifying statics surely?

gausie (approved):
This is neat!

src/mood.ts
phulin: yeah, will tweak this in the next PR. I want to be able to set defaults for a whole script, e.g. for a script in-run to not use sausages in any of its moods.

PR #24, opened 2021-01-29
Publish documentation

gausie:
This should already be [live](https://loathing-associates-scripting-society.github.io/libram/).

I also renamed (and slightly refactored) `banishedMonsters` to `getBanishedMonsters` and put it in `lib.ts` instead of `combat.ts` (@phulin)

Also adds a README


gausie:
Maybe I should change this to be on release instead of on master push

pastelmind:
Awesime work.

> Maybe I should change this to be on release instead of on master push

Ideally we

pastelmind (approved):

PR #28, opened 2021-02-10
style: Various ESLint/Prettier/TypeScript fixes

pastelmind:
- Ran `yarn run format` to prettify code
- Ran `yarn run lint` to detect errors
- Replaced some `eslint-disable` comments with better fixes
- Use `unknown` instead of `any` type
   (This requires adding multiple @ts-expect-error comments in `property.ts` just to make things compile. IMO it shows that our types have flaws that should be fixed in a more dedicated PR)

Note: `husky` auto-formatted `propertyTyping.ts` and broke the extremely long lines into multiple short lines. No complaints from me on this!


gausie:
I don't like the bits of this PR that mess with the properties code. I don't like breaking up those lines as they are autogenerated and I think the `unknown` is not necessary there

PR #32, opened 2021-02-23
Add some more utility functions

gausie:
For parsing mafia adventure ranges, checking if items are usable in the current path etc


src/lib.ts
```diff
@@ -396,4 +397,76 @@ export function getBanishedMonsters(): Map<Item | Skill, Monster> {
...
+/**
+ * Return average adventures expected from consuming an item
+ *
+ * If item is not a consumable, will just return "0".
+ *
+ * @param item Consumable item
+ */
+export function getAverageAdventures(item: Item): number {
+  return getAverage(item.adventures);
 }
```
phulin, line 472: did prettier run? needs a newline here.

phulin (approved):

src/lib.ts
gausie: Just ran it again and it's not adding one...

PR #33, opened 2021-02-27
Polyfill Object.values

gausie:
Object.values isn't supplied by Rhino. I thought that babel would convert it but it seems you need to polyfill. Anyway so this should fix it without importing every single polyfill in the world.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "webpack-cli": "^4.2.0"
   },
   "dependencies": {
+    "core-js": "3.9.0",
     "kolmafia": "^1.0.3",
     "node-html-parser": "^2.0.0"
   },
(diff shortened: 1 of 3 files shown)


phulin:
Did you try preset-env with core-js: usage in the config

phulin:
never mind

phulin (approved):

pastelmind:
Does this violate the ["no side effects" declaration in `package.json`? Also, we might want to think about downstream code that may want to use their own polyfill for `Object.values()`.

https://github.com/Loathing-Associates-Scripting-Society/libram/blob/bcf51e34bb722c164e14910599b84e6c55c50496/package.json#L62

PR #37, opened 2021-03-10
Miscellaneous bugfixes.

phulin:


phulin:
@gausie can you review

src/resources/2018/SongBoom.ts
```diff
@@ -24,6 +24,14 @@ export const songBoomSongs = new Set<string>([
   "Total Eclipse of Your Meat",
 ]);
 
+const keywords = {
```
gausie, line 27: Can we stop these three repeated sections and just say

```ts
type SongBoomSong = keyof keywords | null;
```
and replace things like `songBoomSongs.has(x)` with `x in keywords`

In doing so I would rename `keywords` to `songBoomSongs`

gausie (changes requested):
Just one change

src/resources/2018/SongBoom.ts
phulin: The keywords don't need to be exposed. Otherwise done.

gausie (approved):

PR #41, opened 2021-05-24
Update props, add familiar and stat prop types to get too

gausie:


gausie:
Did this again in #45

PR #42, opened 2021-05-24
Mafia now expects a function as a combat filter instead of a string

gausie:


Phillammon (approved):

gausie:
it takes both, we don't need this

PR #48, opened 2021-07-16
Preliminary stash support

gausie:


src/Clan.ts
```diff
@@ -245,4 +280,67 @@ export class Clan {
...
+  withStash<T>(items: Item[], callback: (borrowedItems: Item[]) => T): T {
+    const borrowed = this.take(items);
+    try {
+      return callback(borrowed);
+    } finally {
+      const returnedItems = this.put(borrowed);
+      const diff = difference(borrowed, returnedItems);
+      if (diff.length > 0) {
+        // eslint-disable-next-line no-unsafe-finally
+        throw new Error(`Failed to return ${diff} to ${this.name}`);
```
phulin, line 341: I don't think we should throw errors in the finally clause. Can we print in a different color instead? That way all items you can't return will be shown.

src/Clan.ts
```diff
@@ -245,4 +280,67 @@ export class Clan {
...
+   * Return the monster that is currently in the current clan's fax machine if any
+   */
+  @validate
+  withStash<T>(items: Item[], callback: (borrowedItems: Item[]) => T): T {
+    const borrowed = this.take(items);
+    try {
+      return callback(borrowed);
+    } finally {
+      const returnedItems = this.put(borrowed);
+      const diff = difference(borrowed, returnedItems);
```
phulin, line 338: I know this is validated based on Mafia's clan id, but I'd like to add a specific visitUrl call to the clan stash page as the Mafia tracking has gotten messed up a few times. Maybe check if the clan name is in the page?

src/Clan.ts
```diff
@@ -78,15 +80,44 @@ export class Clan {
...
+   * During the execution of the callback, player will not be in the stash clan
+   * 
+   * @param clanIdOrName Clan id or name
+   */
+  static withStash<T>(clanIdOrName: string | number, items: Item[], callback: (borrowedItems: Item[]) => T): T {
+    const borrowed = Clan.with(clanIdOrName, clan => clan.take(items));
+    try {
+      return callback(borrowed);
+    } finally {
+      const returnedItems = Clan.with(clanIdOrName, clan => clan.put(borrowed));
```
phulin, line 114: This should only change clans if we have taken stuff out - if nothing was available we shouldn't change.

pastelmind:
It's a pity that JS does not support chaining exceptions natively, which makes context managers less robust than we would like them to be (i.e. not being able to throw inside a `finally` block). We could provide a custom exception chaining system by catching, decorating, and re-throwing exceptions. However, rethrowing an exception causes us to lose the original stack trace, making it difficult to debug such errors.

What are your thoughts? Would we benefit from something like the `Either` type, popularized by FP languages?

src/Clan.ts
gausie: I feel like this is a catastrophic failure so I erred on throwing - I don't actually have a problem with throwing here. I would rather not print from libram, want to leave that to the consumer. What do you think?

src/Clan.ts
gausie: Great point!

horrible-little-slime:
How easy would it be to add the ability to yoink multiples of the same item?

src/Clan.ts
```diff
@@ -245,4 +298,73 @@ export class Clan {
...
+      if (foldGroup.some(fold => have(fold))) {
+        cliExecute(`fold ${item.name}`);
+        return false;
+      }
+
+      refreshStash();
+
+      return [item, ...foldGroup].some((matchingItem) => {
+        if (stashAmount(matchingItem) === 0) return false;
+        if (!takeStash(1, matchingItem)) return false;
```
horrible-little-slime, line 326: If `takeStash` fails for one item in a fold group, do we expect it to succeed for any others? If a fold group is 0-karma, the entire fold group is 0-karma. 

Similarly, is there value in drawing a distinction between items you already had and items you failed to yoink in `take()`?

src/Clan.ts
```diff
@@ -245,4 +298,73 @@ export class Clan {
...
+   */
+  @validate
+  take(items: Item[]): Item[] {
+    return items.filter((item) => {
+      if (have(item)) return false;
+
+      const foldGroup = getFoldGroup(item);
+
+      if (foldGroup.some(fold => have(fold))) {
+        cliExecute(`fold ${item.name}`);
```
horrible-little-slime, line 318: I know this part comes straight from me, but is there a nicer way to achieve this than cliExecute?

src/Clan.ts
```diff
@@ -245,4 +298,73 @@ export class Clan {
...
+  /**
+   * Take items from the stash
+   * 
+   * This function will also take equivalent foldables if the original item cannot be found
+   * 
+   * @param items Items to take
+   * @returns Items successfully taken
+   */
+  @validate
+  take(items: Item[]): Item[] {
```
horrible-little-slime, line 311: Is there a non-terrible way to set this up to let us take multiple of an item/fold group? Right now, because we skip if we have it, putting the same item in the array multiple times does nothing.

src/Clan.ts
gausie: Nope!

src/Clan.ts
gausie: Yeah I noticed that, it's also an issue with the garbo implementation. I can definitely do it as a map or entries-style array

src/Clan.ts
gausie: > If `takeStash` fails for one item in a fold group, do we expect it to succeed for any others? If a fold group is 0-karma, the entire fold group is 0-karma.
Probably not but I'm inclined to keep the code simple.

> Similarly, is there value in drawing a distinction between items you already had and items you failed to yoink in `take()`?
No I don't think so - the only thing we really care about is what we took so we can return it after

gausie:
> It's a pity that JS does not support chaining exceptions natively, which makes context managers less robust than we would like them to be (i.e. not being able to throw inside a `finally` block). We could provide a custom exception chaining system by catching, decorating, and re-throwing exceptions. However, rethrowing an exception causes us to lose the original stack trace, making it difficult to debug such errors.
> 
> What are your thoughts? Would we benefit from something like the `Either` type, popularized by FP languages?

@pastelmind I completely agree and would love an `Either` type. I reckon it could get confusing here though

horrible-little-slime (commented):
Looks generally good--I really appreciate the option of pulling more than one of the same item.

PR #49, opened 2021-07-18
use get, not post when abandoning a guzzlr uest

horrible-little-slime:


horrible-little-slime:
@gausie I don't think I can request a review, but this is me requesting a review

gausie (approved):

PR #50, opened 2021-07-18
Switch lodash import format.

phulin:
This reduces bundle sizes of importers of the library substantially.


gausie:
This is technically a module consumer issue - libram doesn't provide a bundled build. You could use something like https://github.com/lodash/babel-plugin-lodash if you wanted. That said, I'll merge this.

PR #52, opened 2021-07-19
Kmail support

gausie:


src/Kmail.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    return Number(results.match(/<td>(\d) messages? deleted.<\/td>/)?.[1] ?? 0);
+  }
+
+  private static _genericSend(
+    to: string | number,
+    message: string,
+    items: Map<Item, number> | Item[],
+    meat: number,
+    chunkSize: number,
```
horrible-little-slime, line 70: is it worth setting chunkSize's default value to 11 and/or verifying that you don't put a chunk size greater than 11 in there?

PR #54, opened 2021-07-21
rudimentary support for gash-hopping

horrible-little-slime:
Does not currently handle post-ascension choice adventures, or perming skills. Does not currently encode much information into the Path class. Throws when slightly inconvenienced, which is probably good.

### DIFF
--- a/src/Path.ts
+++ b/src/Path.ts
@@ -1,0 +1,62 @@
+export class Path {
+  name: string;
+  id: number;
+  isAvatar: boolean; //here, we define avatar-ness around being its own class
+  constructor(name: string, id: number, isAvatar = false) {
+    this.name = name;
+    this.id = id;
+    this.isAvatar = isAvatar;
+  }
+
+  static unrestricted: Path = new Path("Unrestricted", 0);
+  static boozetafarian: Path = new Path("Boozetafarian", 1);
+  static teetotaler: Path = new Path("Teetotaler", 2);
+  static oxygenarian: Path = new Path("Oxygenarian", 3);
+  static beesHateYou: Path = new Path("Bees Hate You", 4);
+  static wayOfTheSurprisingFist: Path = new Path(
+    "Way of the Surprising Fist",
+    6
+  );
+  static trendy: Path = new Path("Trendy", 6);
+  static avatarOfBoris: Path = new Path("Avatar of Boris", 8, true);
+  static bugbearInvasion: Path = new Path("Bugbear Invasion", 9);
+  static zombieSlayer: Path = new Path("Zombie Slayer", 10, true);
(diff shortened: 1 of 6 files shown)


src/ascend.ts
```diff
@@ -0,0 +1,292 @@
...
+  print,
+  toInt,
+  use,
+  visitUrl,
+  xpath,
+} from "kolmafia";
+import { $item, $items, $stat, get, have } from ".";
+import { ChateauMantegna } from "./resources";
+
+export enum lifestyle {
```
gausie, line 15: capitalise

src/ascend.ts
```diff
@@ -0,0 +1,292 @@
...
+  casual = 1,
+  softcore = 2,
+  normal = 2,
+  hardcore = 3,
+}
+
+export class Path {
```
gausie, line 22: put all the path stuff into a Path.ts
combine the list into the class itself as static

src/ascend.ts
```diff
@@ -0,0 +1,292 @@
...
+        case $stat`Mysticality`:
+          return 1;
+        case $stat`Moxie`:
+          return 2;
+        default:
+          throw `unknown prime stat for ${playerClass}`;
+      }
+    };
+
+    switch ((moon as string).toLowerCase()) {
```
gausie, line 109: Pull this out into some sort of moon sign function

gausie (changes requested):

src/Path.ts
```diff
@@ -0,0 +1,62 @@
...
+    true
+  );
+  static oneCrazyRandomSummer: Path = new Path("One Crazy Random Summer", 24);
+  static communityService: Path = new Path("Community Service", 25);
+  static avatarOfWestOfLoathing: Path = new Path(
+    "Avatar of West of Loathing",
+    26,
+    true
+  );
+  static theSource: Path = new Path("The Source", 27);
```
gausie, line 46: Title case these and add a `Path.all()` functiont to get all of them for looping and I think we're done

horrible-little-slime: per our conversation in discord, I am not doing this and am instead making a Paths object again

src/Path.ts
```diff
@@ -0,0 +1,176 @@
...
+    this.hasCampground = hasCampground;
+    this.hasTerrarium = hasTerrarium;
+    this.stomachSize = stomachSize;
+    this.liverSize = liverSize;
+    this.spleenSize = spleenSize;
+    this.classes = classes;
+  }
+}
+
+export const Paths = {
```
gausie, line 36: You might need to `as const` this

gausie (approved):

src/Path.ts
horrible-little-slime: done!

PR #59, opened 2021-07-25
Rewrite maximize cache to store equipment list and multiple entries

Rinn:
- Keep track of all previous maximize requests and the resulting equipment
- If a previous cache entry is found, simply re-equip those items in the correct slot. If the equipment fails to be put in the correct slot, fallback and maximize again
- Persist setting to optionally invalidate cache if familiar changes
- Change updateOnStatThreshold setting to updateOnCanEquipChanged, which compares the total number of uniquely items that could be equipped, which prevents unnecessary maximization if stats change but no new items could be worn. This number is also cached so the item list isn't filtered again if stats are unchanged.
- locale sort the maximize string by group, to ensure consistent ordering within the string
- Round bonus number to two decimal places
- Add proper variables for +slot (only consider this slot) & -slot (ignore this slot entirely), don't cache slots that aren't appropriate if those options are used. -slot isn't passed for buddy-bjorn or crown-of-thrones because ignoring those slots is kind of busted in mafia, however those slots will not be cached.

### DIFF
--- a/src/maximize.ts
+++ b/src/maximize.ts
@@ -170,7 +170,10 @@ function verifyCached(entry: CacheEntry): boolean {
     }
   }
 
-  if (equippedAmount($item`Crown of Thrones`) > 0) {
+  if (
+    equippedAmount($item`Crown of Thrones`) > 0 &&
+    entry.rider.get($item`Crown of Thrones`)
+  ) {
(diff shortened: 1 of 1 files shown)


gausie (approved):
This looks good to me. If it works, happy to accept it

PR #60, opened 2021-07-28
lastVoteMonsterTurn is a numeric property

s-k-z:
Moved this property into the other category. One thing I'm unsure about though is that since the default is an empty string, should this be a `NumericOrStringProperty` instead?


gausie:
> Moved this property into the other category. One thing I'm unsure about though is that since the default is an empty string, should this be a `NumericOrStringProperty` instead?

We want to make sure that blank strings get returned as 0. Not sure if that's already the case

PR #61, opened 2021-08-02
create PropertyManager class, to avoid frequent flip-flopping with withChoice and withProperty wrappers

horrible-little-slime:
per discussion with @phulin in discord

### DIFF
--- a/src/property.ts
+++ b/src/property.ts
@@ -210,10 +210,8 @@ export function withChoice(
 
 export class PropertiesManager {
   properties: Properties;
-  choices: { [choice: number]: number | string };
   constructor() {
     this.properties = {};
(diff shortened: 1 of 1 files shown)


src/property.ts
```diff
@@ -172,17 +179,63 @@ export function withProperties(properties: Properties, callback: () => void): vo
...
+  callback: () => void
+): void {
   withChoices({ [choice]: value }, callback);
-}
\ No newline at end of file
+}
+
+export class PropertiesManager {
+  properties: Properties;
+  choices: { [choice: number]: number | string };
```
phulin, line 213: Keep this as one set of properties I think

gausie (approved):

PR #64, opened 2021-08-06
Uncircularize Imports

horrible-little-slime:
I'm not actually sure if this is where things went awry, but either ascend.ts or ChateauMantegna.ts introduced circular imports with its use of template strings. So let's see if this is any better.

### DIFF
--- a/src/Path.ts
+++ b/src/Path.ts
@@ -1,4 +1,4 @@
-import { $classes } from ".";
+import { $classes } from "./template-string";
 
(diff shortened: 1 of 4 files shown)


horrible-little-slime:
@gausie @phulin

phulin:
Don’t @ me

phulin:
(That was a joke)

phulin (approved):

PR #65, opened 2021-08-07
make the propertiesmanager store properties correctly

horrible-little-slime:
At present, the PropertiesManager doesn't store properties properly. I would like it to do so.

### DIFF
--- a/src/property.ts
+++ b/src/property.ts
@@ -218,7 +218,7 @@ export class PropertiesManager {
     for (const [propertyName, propertyValue] of Object.entries(
       propertiesToSet
     )) {
-      if (this.properties[propertyName as KnownProperty] !== undefined) {
+      if (this.properties[propertyName as KnownProperty] === undefined) {
         this.properties[propertyName as KnownProperty] = get(propertyName);
       }
       set(propertyName, propertyValue);
--- a/src/property.ts
+++ b/src/property.ts
@@ -226,13 +226,14 @@ export class PropertiesManager {
   }
 
   setChoices(choicesToSet: { [choice: number]: number | string }): void {
-    Object.entries(choicesToSet).forEach(([choiceNumber, choiceValue]) => {
-      const propertyName = `choiceAdventure${choiceNumber}`;
-      if (!((propertyName as KnownProperty) in this.properties)) {
-        this.properties[propertyName as KnownProperty] = get(propertyName);
-      }
-      set(propertyName, choiceValue);
-    });
+    this.set(
+      Object.fromEntries(
+        Object.entries(choicesToSet).map(([choiceNumber, choiceValue]) => [
+          `choiceAdventure${choiceNumber}` as KnownProperty,
+          choiceValue,
+        ])
+      )
+    );
   }
 
   resetAll(): void {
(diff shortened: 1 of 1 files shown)


src/property.ts
```diff
@@ -215,27 +215,22 @@ export class PropertiesManager {
...
 
   set(propertiesToSet: Properties): void {
-    this.properties = {
-      ...Object.keys(propertiesToSet).map((propertyName) => [
-        propertyName,
-        get(propertyName),
-      ]),
-      ...this.properties,
-    };
     Object.entries(propertiesToSet).forEach(([propertyName, propertyValue]) => {
```
phulin, line 225: Can you use for...of instead of forEach? Better style.

src/property.ts
```diff
@@ -215,27 +215,22 @@ export class PropertiesManager {
...
   setChoices(choicesToSet: { [choice: number]: number | string }): void {
-    this.set(
-      Object.fromEntries(
-        Object.entries(choicesToSet).map(([choiceNumber, choiceValue]) => [
-          `choiceAdventure${choiceNumber}` as NumericOrStringProperty,
-          choiceValue,
-        ])
-      )
-    );
+    Object.entries(choicesToSet).forEach(([choiceNumber, choiceValue]) => {
```
phulin, line 227: Why not use this.set? DRY

src/property.ts
```diff
@@ -215,27 +215,22 @@ export class PropertiesManager {
...
   set(propertiesToSet: Properties): void {
-    this.properties = {
-      ...Object.keys(propertiesToSet).map((propertyName) => [
-        propertyName,
-        get(propertyName),
-      ]),
-      ...this.properties,
-    };
     Object.entries(propertiesToSet).forEach(([propertyName, propertyValue]) => {
+      if (!((propertyName as KnownProperty) in this.properties)) {
```
phulin, line 219: I prefer o[k] === undefined for this pattern, and I think that's what we use elsewhere.

phulin (changes requested):

src/property.ts
horrible-little-slime: I originally did that, but my concern was that I wasn't sure the object I was constructing in the way I set it up would actually act as a Properties thing. I've changed it to use this.set, but make sure to give that section a careful look to make sure it's not just entirely wrong

src/property.ts
```diff
@@ -215,23 +215,21 @@ export class PropertiesManager {
...
-        propertyName,
-        get(propertyName),
-      ]),
-      ...this.properties,
-    };
-    Object.entries(propertiesToSet).forEach(([propertyName, propertyValue]) => {
+    for (const [propertyName, propertyValue] of Object.entries(
+      propertiesToSet
+    )) {
+      if (this.properties[propertyName as KnownProperty] !== undefined) {
```
phulin, line 221: shouldn't this only be if it DOES equal undefined? I.e. save a property's value if it's the first time we are setting it in the session?

phulin (changes requested):

src/property.ts
horrible-little-slime: yep, force of habit. Whoops!

phulin (approved):

PR #68, opened 2021-08-07
build: use import helpers from tslib

pastelmind:
Previously, the TypeScript compiler added helper functions (e.g. `__extend()`, `__makeTemplateObject()`, `__values()`) separately to each emitted `.js` file. This caused a non-trivial amount of code duplication when libram was bundled by a downstream project.

For example, see https://github.com/Loathing-Associates-Scripting-Society/garbage-collector/blob/99e5c66fe702290eb6dab8f6d58adf8ae40087d2/scripts/garbage-collector/garbo.js and count the number of `var __makeTemplateObject` statements.

To solve this, this PR enables `compilerOptions.importHelpers`, which makes TypeScript import the helper functions from [tslib](https://github.com/microsoft/tslib). This also adds tslib to our `dependencies` in `package.json` so that downstream projects can pick it up automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
   "dependencies": {
     "core-js": "3.15.2",
     "kolmafia": "^1.1.2",
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "tslib": "^2.3.0"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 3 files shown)


phulin (approved):
Good find!

PR #69, opened 2021-08-08
only set&save properties we intend to actually change in propertiesmanager.set

horrible-little-slime:


gausie:
This would mean that

```ts
function weirdBehaviour() {
  const pm = new PropertiesManager();
  pm.set({ a: 1 });
  set("a", 5);
  pm.resetAll();
  return get("a");
}
```

would return 1 if `get("a") === 1` before running it or 5 if `get("a") !== 1` before running it

horrible-little-slime:
Closed per Gausie's remark; upon further reflection, it's better to redundantly reset properties than it is to fail to do so.

PR #70, opened 2021-08-08
build: build with Babel instead of TypeScript

pastelmind:
Currently, we explicitly import `core-js` polyfills in several places so that we can use features unsupported by Rhino (e.g. `Array.prototype.flatMap()`). These polyfills add side effects which prevent downstream bundlers from tree-shaking libram correctly. It also forces downstream projects to use polyfills that we choose, rather than giving them a choice.

To solve this problem, this PR uses Babel CLI as a build tool so that we can emit "pure" polyfill code that doesn't affect globals.

Detailed changes:

1. Update `@babel/*` packages to align with Babel 7.15.0, which supports Rhino as a build target (thanks to @gausie).
    1. Since we no longer need to monkey-patch `@babel/compat-data` and `@babel/helper-compilation-targets`, let's remove `patch-package` as well as the `*.patch` files.
2. Add `@babel/plugin-transform-runtime` and `@babel/runtime-corejs3` so that...
    1. Helper functions are imported from `@babel/runtime-corejs3`, which allows bundlers to deduplicate them (i.e. what `tslib` used to do).
    2. Use polyfills without polluting global scope, and keeping our code free of side effects.
3. Use the TypeScript compiler only to generate `.d.ts` files. Update `tsconfig.json` accordingly.
    1. Remove `tslib` since `@babel/runtime-corejs3` now does its job.

---

However, there's a major problem with this PR: Babel tries to polyfill _every_ ES6+ feature, even those that Rhino already supports. For example, all uses of `Set` and `Map` are replaced with implementations provided by `@babel/runtime-corejs3`.

To reproduce this, do:

1. `git checkout` this branch
2. Run `yarn install --frozen-lockfile`
3. Run `yarn run build`
4. Open `dist/Kmail.js` and search for expressions that look like `new _Map(...)`

I'm lost as to how to fix this in Babel.

---

One alternative would be to rewrite _our_ code to use ponyfills instead of polyfills. It makes our DX slightly more tedious, but would be much easier than trying to fix `@babel/plugin-transform-runtime`.


pastelmind:
Also, Babel 7.15.0 transpiles Rhino-supported syntax (e.g. arrow functions, `for...of` loops) to ES5, even with `"targets": { "rhino": "1.7" }` in `babel.config.json`. Is my configuration incorrect?

phulin:
Maybe set rhino version to 1.7.13?

gausie:
I noticed this issue with core-js too. I think for now we could keep explicitly importing polyfills? If it's transpiling a feature we have it's probably because that feature fails one of the unit tests and is thus not fully implemented by rhino

pastelmind:
> Maybe set rhino version to 1.7.13?

Tried this and it works--arrow functions are now preserved. Unfortunately, `for...of` loops are still transpiled. I assume Rhino's iterator implementation may be failing some conformance tests.

gausie (approved):

PR #73, opened 2021-08-12
add Requirement from garbo

horrible-little-slime:


gausie:
Use a leading `_` instead of a trailing `_` for the private values, or better yet use the [new private syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields). Then you could maybe use a getter for the public method.

src/maximize.ts
```diff
@@ -329,3 +329,65 @@ export function maximizeCached(
...
+    return this.maximizeOptions_;
+  }
+
+  merge(other: Requirement): Requirement {
+    const optionsA = this.maximizeOptions();
+    const optionsB = other.maximizeOptions();
+    return new Requirement(
+      [...this.maximizeParameters(), ...other.maximizeParameters()],
+      {
+        ...optionsA,
```
gausie, line 359: Can we not just use deepMerge from lodash here?

horrible-little-slime: Changed; I think I've done what you're looking for, but I'm not 100% sure.

src/maximize.ts
```diff
@@ -329,3 +330,40 @@ export function maximizeCached(
...
+export class Requirement {
+  #maximizeParameters: string[];
+  #maximizeOptions: MaximizeOptions;
+
+  constructor(maximizeParameters: string[], maximizeOptions: MaximizeOptions) {
+    this.#maximizeParameters = maximizeParameters;
+    this.#maximizeOptions = maximizeOptions;
+  }
+
+  maximizeParameters(): string[] {
```
gausie, line 343:
```suggestion
  get maximizeParameters(): string[] {
```

src/maximize.ts
```diff
@@ -329,3 +330,40 @@ export function maximizeCached(
...
+    return this.#maximizeParameters;
+  }
+
+  maximizeOptions(): MaximizeOptions {
```
gausie, line 347:
```suggestion
  get maximizeOptions(): MaximizeOptions {
```

src/maximize.ts
```diff
@@ -14,6 +14,7 @@ import {
 } from "kolmafia";
 import { $familiar, $item, $slot, $slots, $stats } from "./template-string";
 import logger from "./logger";
+import _ from "lodash";
```
gausie, line 17:
```suggestion
import merge from "lodash/merge";
```

src/maximize.ts
```diff
@@ -329,3 +330,40 @@ export function maximizeCached(
...
+  maximizeParameters(): string[] {
+    return this.#maximizeParameters;
+  }
+
+  maximizeOptions(): MaximizeOptions {
+    return this.#maximizeOptions;
+  }
+
+  merge(other: Requirement): Requirement {
+    return _.merge(this, other);
```
gausie, line 352:
```suggestion
    return merge(this, other);
```

gausie (changes requested):

PR #76, opened 2021-08-15
florist friar support

horrible-little-slime:


src/resources/2013/Florist.ts
```diff
@@ -0,0 +1,364 @@
...
+  location: Location = myLocation()
+): Flower[] {
+  return all().filter((flower) => flower.available(location));
+}
+
+export function isFull(location = myLocation()): boolean {
+  return flowersIn(location).length == 2;
+}
+
+export const all: () => Flower[] = () => [
```
gausie, line 91: Why is this a function?

src/resources/2013/Florist.ts
```diff
@@ -0,0 +1,364 @@
...
+  ) {
+    this.name = name;
+    this.id = id;
+    this.environment = environment;
+    this.modifier = modifier;
+    this.territorial = territorial;
+  }
+
+  isPlantedHere(location = myLocation()): boolean {
+    return getFloristPlants()[location.toString()].includes(this.name);
```
gausie, line 33: This needs to have `?.`

src/resources/2013/Florist.ts
```diff
@@ -0,0 +1,364 @@
+import {
+  floristAvailable,
+  getFloristPlants,
+  myLocation,
+  visitUrl,
+} from "kolmafia";
+import { get } from "../../property";
+
+type environment = "outdoor" | "indoor" | "underground" | "underwater";
```
gausie, line 9: imo we add

```ts
export Environment = {
  Outdoor: "outdoor",
  Indoor: "indoor",
  Underground: "underground",
  Underwater: "underwater",
} as const;

export type EnvironmentType = typeof Environment[keyof typeof Environment];
```
to `lib.ts`, and we type environment here as `EnvironmentType`

src/resources/2013/Florist.ts
```diff
@@ -0,0 +1,364 @@
...
+      !get("_floristPlantsUsed").includes(this.name) &&
+      !this.isPlantedHere(location)
+    );
+  }
+
+  dig(): boolean {
+    if (!this.isPlantedHere()) return false;
+    const flowers = getFloristPlants()[myLocation().toString()];
+    if (!flowers[2]) return false;
+    const plantNumber = getFloristPlants()[myLocation().toString()].indexOf(
```
gausie, line 48: All of this `getFloristPlants()[myLocation().toString()].whatever` probably needs to be extracted out into a shared function(s) to handle undefineds

src/resources/2013/Florist.ts
horrible-little-slime: Part if it was to prevent mutation, and part of it was because writing `FloristFriar.all()` felt natural. But yeah, there's no actual compelling reason to do so.

gausie: Use `Object.freeze` to prevent mutation, and if you want a function thats fine but probably return an externally defined array!

src/resources/2013/Florist.ts
horrible-little-slime: This one specifically can stay, but the others are extracted

gausie:
@horrible-little-slime are you wanting to mark this as ready for review?

horrible-little-slime:
Sure! I'm not sure if this is actually ready, but I think I'm done working
on it.

gausie (approved):
I say we merge it and wait for bug reports from users I guess!

PR #87, opened 2021-08-20
commit questStep function

JamesDowney:
questStep is a function with a lineage. I stole it from Garbo, which I think may have originally gotten it from Rever. I know a lot of scripts use a variation of questStep, and I plan on using it in a future canadv knockoff, so I figured this is a good time to bake it into Libram.

The reasoning behind it returning the same value for an empty string as a finished quest is, as Phredd pointed out to me, that the intro quest at NEP will set the corresponding property to an empty string if it is rejected.


src/lib.ts
```diff
@@ -33,6 +33,7 @@ import {
 import { $class, $items } from "./template-string";
 import { get } from "./property";
 import { chunk } from "./utils";
+import { property } from ".";
```
phulin, line 36:
```suggestion
import * as property from "./property";
```

gausie (approved):

PR #92, opened 2021-08-29
Bugbear invasion property change

JamesDowney:
Moving bugbear invasion properties away from being numeric properties to numeric or string properties to be more in line with mafia's usage.

For reference:
https://wiki.kolmafia.us/index.php/Quest_Tracking_Preferences#Bug_Bear_Invasion_Properties


gausie:
Thank you! For this to persist we also need to update isNumericOrStringProperty, which I have done in 7d6348, superseding this PR.

PR #94, opened 2021-09-01
Add outfit caching.

phulin:
This adds outfit caching to enable faster outfit switching.

I did take the opportunity to add `Partial` typing to the MaximizeOptions object which is a slight breaking change for garbo.


gausie (approved):
LGTM

PR #95, opened 2021-09-01
add beach head support

horrible-little-slime:


src/lib.ts
```diff
@@ -506,3 +507,18 @@ export function getPlayerFromIdOrName(idOrName: number | string): Player {
...
+
+/**
+ * Tries to get an effect using the default method
+ * @param ef effect to try to get
+ * @param turns turns to aim for; default of 1
+ */
+export function ensureEffect(ef: Effect, turns = 1): void {
+  if (haveEffect(ef) < turns) {
+    if (!cliExecute(ef.default) || haveEffect(ef) === 0) {
+      throw `Failed to get effect ${ef.name}.`;
```
gausie, line 519: I understand why throwing here is necessary. Let's make an `EnsureError` or something, and throw that. And make sure its exported by libram too.

src/lib.ts
```diff
@@ -506,3 +507,18 @@ export function getPlayerFromIdOrName(idOrName: number | string): Player {
...
+    }
+  } else {
+    print(`Already have effect ${ef.name}.`);
```
gausie, line 522: Don't print from libram

src/resources/2019/BeachComb.ts
```diff
@@ -0,0 +1,25 @@
+import { ensureEffect } from "../..";
+import { getString } from "../../property";
+import { $effect } from "../../template-string";
+
+const heads = [
```
gausie, line 5: can you export this? And call it `buffs` or something. And `as const` it.

src/resources/2019/BeachComb.ts
```diff
@@ -0,0 +1,25 @@
...
+  $effect`Oiled, Slick`,
+  $effect`Lack of Body-Building`,
+  $effect`We're All Made of Starfish`,
+  $effect`Pomp & Circumsands`,
+  $effect`Resting Beach Face`,
+  $effect`Do I Know You From Somewhere?`,
+  $effect`You Learned Something Maybe!`,
+];
+
+export function getHead(effect: Effect): void {
```
gausie, line 19: Sadly we should just call this `getBuff`

src/resources/2019/BeachComb.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (!heads.includes(effect)) return;
+  const headNumber = 1 + heads.indexOf(effect);
+  if (getString("_beachHeadsUsed").split(",").includes(headNumber.toString()))
+    return;
+  ensureEffect(effect);
```
gausie, line 24: Why not make this function return a success boolean?

gausie (changes requested):

src/resources/2019/BeachComb.ts
```diff
@@ -0,0 +1,26 @@
...
+  $effect`Do I Know You From Somewhere?`,
+  $effect`You Learned Something Maybe!`,
+] as const;
+
+export function tryHead(effect: Effect): boolean {
+  if (!headBuffs.includes(effect)) return false;
+  const headNumber = 1 + headBuffs.indexOf(effect);
+  if (getString("_beachHeadsUsed").split(",").includes(headNumber.toString()))
+    return false;
+  ensureEffect(effect);
```
gausie, line 24: If we're doing ensureEffect there's no point returning boolean `have()`. I would remove the ensureEffect

gausie (approved):

PR #108, opened 2021-09-12
Knapsack solver for diet planning.

phulin:
This PR adds a knapsack-based estimator to plan diets. It supports Refined Palate but not yet Gar-ish (a significant issue for now, but trying to reduce scope). It will construct a diet for you but will not plan the exact sequence in which to eat it. It also doesn't support Sweet Synthesis yet - an omission we'll have to remedy at some point.

You can test it with the following TS script:

[diet-plan.ts.zip](https://github.com/Loathing-Associates-Scripting-Society/libram/files/7516652/diet-plan.ts.zip)
.


horrible-little-slime:
Account for mushroom wines, and pocket wishes for refined palate; consider accounting for toasted brie for refined palate, but this is likely suboptimal for just about all cases.

src/diet/index.ts
```diff
@@ -0,0 +1,506 @@
...
+  mug?: MenuItem;
+  seasoning?: MenuItem;
+  mayoflex?: MenuItem;
+  pinkyRing: boolean;
+  tuxedoShirt: boolean;
+  spleenValue = 0;
+
+  constructor(mpa: number, menu: MenuItem[]) {
+    this.mpa = mpa;
+    this.fork = menu.find(
```
horrible-little-slime, line 162: It may be wise to not explicitly define these as undefined, because that can make object spread syntax different for reasons that elude me. So you could do something like
```
const mug =  menu.find((item) => item.item === $item`Frosty's frosty mug`);
if (mug) this.mug = mug;
```

src/diet/index.ts
```diff
@@ -0,0 +1,506 @@
...
+      const current = organCapacitiesWithMap.get(organ);
+      if (current !== undefined) {
+        organCapacitiesWithMap.set(organ, current - size);
+      }
+    }
+    const organCapacitiesWith = [...organCapacitiesWithMap];
+
+    const isRefinedPalate =
+      trialItem.item === $item`pocket wish` &&
+      trialItem.wishEffect === $effect`Refined Palate`;
```
horrible-little-slime, line 328: I do not currently see any code that accounts for already having this active, which is something that can plausibly happen when mushroom wine prices are low--you nightcap with a sacramento+mushroom, and then start the day with it. Similarly, it'd be neat if there was a way to account for "having this buff left over is valuable", but that seems hard and I can't identify what precisely I'd want it to do.

horrible-little-slime (commented):
This looks good, but I do not feel comfortable approving it because it is well beyond my understanding. I will revisit this in a few days right after coffee for more suggestions.

src/diet/index.ts
phulin: I don't think this matters, these properties aren't enumerable anyway

src/diet/index.ts
phulin: yep sounds good

gausie (approved):
Let's merge this to get it in people's hands!

PR #109, opened 2021-09-14
Make outift checking more flexible.

phulin:
Don't fail verification if e.g. accessory order has changed.


src/lib.ts
```diff
@@ -569,3 +569,12 @@ export const Environment = {
 } as const;
 
 export type EnvironmentType = typeof Environment[keyof typeof Environment];
+
+export function setEqual<T>(a: T[], b: T[]): boolean {
```
gausie, line 573: Please document this

phulin: 👎

PR #110, opened 2021-09-18
add optional ifFalse input to externalIf macro method, thereby giving us an externalElse

horrible-little-slime:


src/combat.ts
```diff
@@ -271,10 +271,17 @@ export class Macro {
...
-  externalIf(condition: boolean, ifTrue: string | Macro): this {
-    return condition ? this.step(ifTrue) : this;
+  externalIf(
+    condition: boolean,
+    ifTrue: string | Macro,
+    ifFalse?: string | Macro
+  ): this {
+    if (condition) return this.step(ifTrue);
+    if (ifFalse) return this.step(ifFalse);
+    return this;
```
gausie, line 284: I think in this particular instance its clearer what's going on if you did something like this.

```suggestion
    if (condition) {
      return this.step(ifTrue);
    } else {
      if (ifFalse) return this.step(ifFalse);
    }

    return this;
```

gausie (approved):

PR #111, opened 2021-09-19
Bjorn/Crown support (rudimentary)

horrible-little-slime:


horrible-little-slime:
@gausie I plan to add bjorn-picking support after this, but I'd like to get this in before I start doing fancy stuff

@phulin I'm 80% sure you'll be fine with the names I have chosen here

src/modifier.ts
```diff
@@ -97,3 +97,51 @@ export function get(
     return statModifier(subject as Effect, name);
   }
 }
+
+export type BooleanModifierValues = Partial<
```
gausie, line 101: Is there a benefit to having them split via type here? could we not just have

```ts
export type Modifiers = Partial<{
  [modifier in BooleanModifier]: boolean;
  [modifier in ClassModifier]: Class;
  ...
}>;
```

horrible-little-slime:
I think there is likely benefit to suggesting specificity, but mostly I got
red squiggles on my first try

src/modifier.ts
horrible-little-slime: I think there is likely benefit to suggesting specificity, but mostly I got
red squiggles on my first try

horrible-little-slime: I can set it up like properties, though

gausie (approved):

PR #112, opened 2021-09-19
Redirect to choice.php if necessary

s-k-z:
To make these combats consistent with adventureMacro, they should redirect to choice.php when a choice follows a fight. Otherwise, if unhandled, then the user script will crash.

### DIFF
--- a/src/resources/2015/ChateauMantegna.ts
+++ b/src/resources/2015/ChateauMantegna.ts
@@ -1,4 +1,4 @@
-import { buy, getChateau, runCombat, visitUrl } from "kolmafia";
+import { buy, choiceFollowsFight, getChateau, runCombat, visitUrl } from "kolmafia";
 import { $item, $items } from "../../template-string";
 import { get } from "../../property";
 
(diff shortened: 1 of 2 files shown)


src/resources/2015/ChateauMantegna.ts
```diff
@@ -14,9 +14,10 @@ export function paintingFought(): boolean {
   return get("_chateauMonsterFought");
 }
 
-export function fightPainting(): string {
+export function fightPainting(): void {
```
gausie, line 17: Could you not still make sure this returns the string?

src/resources/2016/Witchess.ts
```diff
@@ -44,5 +44,6 @@ export function fightPiece(piece: Monster): string {
   ) {
     throw new Error("Failed to start fight.");
   }
-  return runCombat();
```
gausie, line 47: It looks like really we need a function that looks like

```ts
function ensureRunCombat(): string {
  const result = runCombat();
  if (choiceFollowsFight()) visitUrl("choice.php");
  return result;
}
```

gausie: This doesn't seem resolved to me @s-k-z? Or do I misunderstand

gausie:
@s-k-z if you're happy with this you should mark it as ready for review! Also I still think there is some code duplication, I unresolved one of my comments

s-k-z:
If I export a shared function for both of these functions to rely on, there's no way to make it not exported as part of Libram's API to users is there? Ideally I'd like to remove the code duplication but since these two functions exist in separate places I'm not sure how to properly do so without exposing some unintended functionality.

gausie:
> If I export a shared function for both of these functions to rely on, there's no way to make it not exported as part of Libram's API to users is there? Ideally I'd like to remove the code duplication but since these two functions exist in separate places I'm not sure how to properly do so without exposing some unintended functionality.

@s-k-z You could make a new `internalUtils.ts` file next to `utils.ts` and write it there

PR #115, opened 2021-09-22
Summon librams

Rinn:
Add functions that return a map of possible results when casting a libram summon skill, taking into account the number of rares already summoned if applicable.

The yellow taffy code could probably be made a bit more elegant.

### DIFF
--- a/src/resources/2007/CandyHearts.ts
+++ b/src/resources/2007/CandyHearts.ts
@@ -1,11 +1,16 @@
-import { $items, $skill } from "../../template-string";
+import { $item, $skill } from "../../template-string";
 import { have as _have } from "../../lib";
 
 export const summonSkill = $skill`Summon Candy Heart`;
-const libramItems = $items`green candy heart, lavender candy heart, orange candy heart, pink candy heart, white candy heart, yellow candy heart`;
-const libramExpected = new Map<Item, number>(
-  libramItems.map((item) => [item, 1.0 / libramItems.length])
-);
+const libramChance = 1.0 / 6;
+const libramExpected = new Map<Item, number>([
+  [$item`green candy heart`, libramChance],
+  [$item`lavender candy heart`, libramChance],
+  [$item`orange candy heart`, libramChance],
+  [$item`pink candy heart`, libramChance],
+  [$item`white candy heart`, libramChance],
+  [$item`yellow candy heart`, libramChance],
+]);
 
 /**
  * @returns true if the player can Summon Candy Heart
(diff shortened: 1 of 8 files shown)


src/resources/2007/CandyHearts.ts
```diff
@@ -0,0 +1,21 @@
...
+export function have(): boolean {
+  return _have(summonSkill);
+}
+
+/**
+ * Returns A map containing the chance of an item to be summoned
+ */
+export function expected(): Map<Item, number> {
+  const results = new Map<Item, number>();
+  libramItems.forEach((item) => results.set(item, 1.0 / libramItems.length));
```
gausie, line 19: I would just construct this Map outside of the file and return it from this function

src/resources/2009/LoveSongs.ts
```diff
@@ -0,0 +1,21 @@
...
+export function have(): boolean {
+  return _have(summonSkill);
+}
+
+/**
+ * Returns A map containing the chance of an item to be summoned
+ */
+export function expected(): Map<Item, number> {
+  const results = new Map<Item, number>();
+  libramItems.forEach((item) => results.set(item, 1.0 / libramItems.length));
```
gausie, line 19: Same as candy hearts

src/resources/2013/PulledTaffy.ts
```diff
@@ -0,0 +1,48 @@
...
+      rareItems.forEach((item) =>
+        results.set(item, rareChance / rareItems.length)
+      );
+    }
+  }
+  commonItems.forEach((item) =>
+    results.set(item, (1.0 - rareChance) / commonItems.length)
+  );
+  return results;
+}
```
gausie, line 48: There's a way of doing this so we don't have the if else nested blocks and it's just a flat sum for each probability. Might be nice to do it that way and also make this (and all of the Maps in this PR) Map using syntax like `return new Map([[k,v],[k,v],[k,v]])`

gausie (commented):
Looks good, thank you so much for writing it. There are a few changes of style I would make. Also check the capitalisation on the doc blocks!

src/resources/2013/PulledTaffy.ts
Rinn: is there a way to merge multiple array.map() and pass them all into the new Map() constructor? if not i'll probably just do this and list out the individual items since there's not very many of them per libram

horrible-little-slime: If you're asking what I think you're asking, you could do ```
array1.map((element) => mappingFunction(element) as [Item, number]);
array2.map((element) => mappingFunction(element) as [Item, number]);
new Map<Item, number>([...array1, ...array2]);
```.

If you aren't asking what I think you're asking, ignore me!

Rinn: i decided to not be overly clever

gausie (approved):

PR #117, opened 2021-09-23
add pickRider function to CrownOfThrones.ts

horrible-little-slime:
I think there is a good 70% chance that this is a terrible way to do this, and that someone has a better idea of how to do this. That being said, there's a 100% chance that this is the only way I could think of to do this. Please, tear this to shreds, this is a learning experience for me.

### DIFF
--- a/src/resources/2010/CrownOfThrones.ts
+++ b/src/resources/2010/CrownOfThrones.ts
@@ -1,4 +1,5 @@
-import { getSaleValue } from "../../lib";
+import { myFamiliar } from "kolmafia";
+import { getSaleValue, have } from "../../lib";
 import { Modifiers } from "../../modifier";
 import { get } from "../../property";
 import { $familiar, $item, $items } from "../../template-string";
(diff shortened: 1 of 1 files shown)


gausie:
What about we lose the cache (I think its just not workable) and do something like

```ts
export function getValidRidingFamiliars(
  includeCurrentFamiliar?: boolean
): FamiliarRider[] {
  return ridingFamiliars.filter(
    (rider) =>
      have(rider.familiar) &&
      !(includeCurrentFamiliar && myFamiliar() === rider.familiar)
  );
}

export function getRiderValue(
  rider: FamiliarRider,
  calculateBonusValue?: (modifiers: Modifiers) => number
): number {
  return (
    (rider.dropPredicate?.() ? rider.probability * rider.meatVal() : 0) +
    (calculateBonusValue?.(rider.modifier) ?? 0)
  );
}

export function pickRider(
  calculateBonusValue: (modifiers: Modifiers) => number = () => 0
): FamiliarRider | null {
  return (
    getValidRidingFamiliars()
      .sort(
        (a, b) =>
          getRiderValue(b, calculateBonusValue) -
          getRiderValue(a, calculateBonusValue)
      )
      .find((rider) => rider.dropPredicate?.()) ?? null
  );
}
```

src/resources/2010/CrownOfThrones.ts
```diff
@@ -575,3 +576,70 @@ export const ridingFamiliars: FamiliarRider[] = [
...
+        )
+    );
+  }
+  const list = riderLists.get(mode);
+  if (list) {
+    while (list[0].dropPredicate && !list[0].dropPredicate()) list.shift();
+    if (!excludeCurrentFamiliar || list[0].familiar !== myFamiliar())
+      return list[0];
+    while (list[1].dropPredicate && !list[1].dropPredicate()) list.splice(1, 1);
+    return list[1];
```
gausie, line 642: This would still be better being based off my example

gausie (approved):

PR #120, opened 2021-09-26
Stop using babel.

phulin:
Babel was preventing tree-shaking in scripts importing libram, causing bundle size to grow dramatically. This change should fix that with no impact on end users.


gausie (approved):
Agreed.

PR #125, opened 2021-10-02
add ifMonster method to Macro class

horrible-little-slime:


gausie:
We should eventually make a polymorphic if statement that calls one of these based on argument type

gausie (approved):

horrible-little-slime:
Bean said something similar. If we're going to shred this upon implementing
it makes sense not to merge, but if we're just redefining, might as well do
so now

PR #127, opened 2021-10-03
use phccs_gash for as an example for ascend.ts

horrible-little-slime:


gausie:
No we don't want this kind of example. This is what we want: https://jsdoc.app/tags-example.html

PR #137, opened 2021-10-21
Added polymorphic macro if

gausie:
Note that I just went ahead and removed `ifMonster` here. I also think this is a good time to `throw`, since I'm basically catching syntax errors.


gausie:
Going to merge this in lieu of a smarter v2

PR #142, opened 2021-11-07
give ifMonster generic typing

horrible-little-slime:


gausie:
This is invalid from the polymorphic if

PR #143, opened 2021-11-08
change useLimitedDrops to ignoreLimitedDrops

horrible-little-slime:
an alternative to this would be to simply remove the ! from valueRider, but I think this is a better solution for reasons


horrible-little-slime:
I should also document all this code, but would like this change to go through (or fail to do so) before I do so

gausie (approved):

PR #146, opened 2021-11-14
IfHolidayWanderer

horrible-little-slime:
create macro predicates that elegantly support holiday wanderers. Currently doesn't handle holiday collisions great, but does handle them a little!


src/lib.ts
```diff
@@ -625,3 +626,23 @@ export function findFairyMultiplier(familiar: Familiar): number {
...
+  ],
+  [
+    "Talk Like a Pirate Day",
+    $monsters`ambulatory pirate, migratory pirate, peripatetic pirate`,
+  ],
+]);
+
+export function getTodaysHolidayWanderers(): Monster[] {
+  const holidays = holiday().split("/");
+  return holidays.map((holiday) => holidayWanderers.get(holiday) ?? []).flat();
```
gausie, line 647:
```suggestion
  return holiday().split("/").flatMap((holiday) => holidayWanderers.get(holiday) ?? []);
```

phulin (approved):

PR #147, opened 2021-11-14
merge defaultMaximizeOptions with inputted options in MaximizeCached

horrible-little-slime:
per discussion on discord; this may be a better behavior, given that this is how garbo was assuming it worked.

### DIFF
--- a/src/maximize.ts
+++ b/src/maximize.ts
@@ -1,4 +1,3 @@
-import merge from "lodash/merge";
 import {
   availableAmount,
   bjornifyFamiliar,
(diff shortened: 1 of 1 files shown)


src/maximize.ts
```diff
@@ -18,6 +18,7 @@ import {
 import { $familiar, $item, $slot, $slots, $stats } from "./template-string";
 import logger from "./logger";
 import { setEqual } from "./utils";
+import merge from "lodash/merge";
```
gausie, line 21: I just merged a `import { maxBy } from "lodash"`. Which one is the correct import style for minimizing bundle size and can we make a change one way or the other to reflect.

Also put this import at the top since its from an external module

horrible-little-slime: If I had to guess, probably importing from lodash/merge and lodash/maxBy would minimize bundle size. Absolutely will move it to the top right now.

horrible-little-slime:
I moved away from lodash's merge because I feel that we want enough properties to behave differently upon merging that it's easier to just hardcode it

phulin:
I think we need to come up with a better solution here. For example, garbo wants to exclude broken champagne bottle from nearly every fight, but there are a couple fights where it wants it. The API should allow you to set that choice and then override it later - this PR doesn't allow removing something from the preventEquip array.

horrible-little-slime:
Fully agreed; this is also a terrible system in general. I do think the
final answer will likely involve ugly hardcoding.

How do you propose we handle overriding defaults?

horrible-little-slime:
Closing this PR; I have a better idea.

PR #151, opened 2021-11-20
Several knapsack bug fixes.

phulin:
* Don't consider organs with capacity < 0.
* Add toasted brie support.
* Add lasgna/Gar-ish support.
* Fix capacity calculation (distention pill/dog hair pill should just be menu items).
* Don't include items with maximum == 0.
* Don't plan out spleen if the user doesn't ask for it.
* Don't include interacting items for normal consideration.

### DIFF
--- a/src/diet/index.ts
+++ b/src/diet/index.ts
@@ -29,8 +29,7 @@ type ConsumptionModifiers = {
   tuxedoShirt: boolean;
 };
 
-// TODO: Include other consumption modifiers - Salty Mouth?
-// TODO: Include Gar-ish etc.
+// TODO: Include Gar-ish and Salty Mouth.
 function expectedAdventures(
   item: Item,
   modifiers: ConsumptionModifiers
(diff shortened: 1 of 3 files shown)


src/diet/index.ts
```diff
@@ -230,9 +251,11 @@ class DietPlanner {
...
       seasoning: this.seasoning ? helpers.includes(this.seasoning) : false,
       mayoflex: this.mayoflex ? helpers.includes(this.mayoflex) : false,
-      refinedPalate: false,
-      pinkyRing: this.pinkyRing,
-      tuxedoShirt: this.tuxedoShirt,
+      refinedPalate: have($effect`Refined Palate`),
+      garish: have($effect`Gar-ish`),
+      saucemaven: have($skill`Saucemaven`),
+      pinkyRing: have($item`mafia pinky ring`),
+      tuxedoShirt: have($item`tuxedo shirt`),
```
horrible-little-slime, line 258: Check for ability to wear a shirt

horrible-little-slime (approved):
I don't see any obvious errors other than the one. The code that I could read I could follow well enough to be confident that it's not broken on the micro scale, but I have enough trouble following on the macro scale that I want to say I have no idea if this will break any larger part of the knapsacking. Still, you're getting a thumbs up from Ebert AND Roeper

PR #153, opened 2021-11-20
pantogram support!

horrible-little-slime:
JSdocs came to a theatre near you

### DIFF
--- a/src/resources/2017/Pantogram.ts
+++ b/src/resources/2017/Pantogram.ts
@@ -153,6 +153,11 @@ const RightSacrifice = new Map<PantogramSacrificeR, [Item | number, number]>([
   ["Moxie Experience Percent", [$item`the funk`, 5]],
 ]);
 
+/**
+ * Finds the item requirements for a particular pair of pants.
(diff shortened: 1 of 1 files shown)


src/resources/2017/Pantogram.ts
```diff
@@ -0,0 +1,224 @@
...
+  | "Meat Drop: 60"
+  | "Item Drop: 15"
+  | "Item Drop: 30"
+  | "Muscle Experience: 3"
+  | "Mysticality Experience: 3"
+  | "Moxie Experience: 3"
+  | "Muscle Experience Percent: 25"
+  | "Mysticality Experience Percent: 25"
+  | "Moxie Experience Percent: 25";
+
```
gausie, line 62: It should be possible to write a bit of TS to extract these from the Alignment, Element tuples, rather than the other way round!

phulin: yeah, should just be `keyof PantogramAlignment`

horrible-little-slime: I don't have a complete image of what you're asking for. Could you please elaborate?

phulin (approved):

PR #156, opened 2021-11-28
allow saving multiple autoattack macros under different names

horrible-little-slime:


horrible-little-slime:
Requesting Manny's review because I feel like he'll have strong opinions on the API. Obviously should still get Gausie or Bean's review before merging.

gausie (approved):
fwiw, code looks good to me

PR #160, opened 2021-12-04
Fix phylum properties and update property list.

phulin:


horrible-little-slime:
Approved! Cant seem to log in on mobile to do so formally.

PR #161, opened 2021-12-04
Change tools to use module syntax.

phulin:
Also updates typescript, ts-node, etc. Tools that run in Node should use a different tsconfig than our code that targets mafia.


gausie:
Fantastic, thanks!

PR #168, opened 2021-12-08
Add support for Dinsey. 

pstalcup:
Does not provide access to all the zone modifiers, but provides all quests with their respective requirements, where the keycards are located, and wrappers to accepting and completing quests

### DIFF
--- a/src/resources/2015/Dinsey.ts
+++ b/src/resources/2015/Dinsey.ts
@@ -40,18 +40,18 @@ export const questPreferenceLocation = new Map<string, Location>([
   ["questEStWorkWithFood", $location`Barf Mountain`],
 ]);
 
-interface questRequirement {
-  item: Item;
+interface QuestRequirement {
+  item: Item | null;
   quantity: number;
   trackingPref?: string;
   completeValue?: number;
 }
 
-export const questRequirements = new Map<string, questRequirement>([
+export const questRequirements = new Map<string, QuestRequirement>([
   [
     "questEStSocialJusticeII",
     {
-      item: $item`none`,
+      item: null,
       quantity: 0,
       trackingPref: "dinseySocialJusticeIIProgress",
       completeValue: 15,
@@ -60,7 +60,7 @@ export const questRequirements = new Map<string, questRequirement>([
   [
     "questEStSocialJusticeI",
     {
-      item: $item`none`,
+      item: null,
       quantity: 0,
       trackingPref: "dinseySocialJusticeIProgress",
       completeValue: 15,
@@ -101,7 +101,7 @@ export const questRequirements = new Map<string, questRequirement>([
   [
     "questEStNastyBears",
     {
-      item: $item`none`,
+      item: null,
       quantity: 0,
       trackingPref: "dinseyNastyBearsDefeated",
       completeValue: 8,
@@ -171,7 +171,7 @@ export function finishQuest() {
   if (currentQuest) {
     const currentQuestRequirement = questRequirements.get(
       currentQuest
-    ) as questRequirement;
+    ) as QuestRequirement;
     if (
       !currentQuestRequirement.trackingPref ||
       getNumber(currentQuestRequirement.trackingPref) ===
--- a/src/resources/2015/Dinsey.ts
+++ b/src/resources/2015/Dinsey.ts
@@ -202,16 +202,20 @@ export function canFightWartDinsey() {
   );
 }
 
-export function disposeGarbage() {
-  if (have($item`bag of park garbage`) && !get("_dinseyGarbageDisposed")) {
-    visitUrl(maintUrl);
-    runChoice(6);
-  }
-}
-
 export function fightWartDinsey() {
   if (canFightWartDinsey()) {
     visitUrl(maintUrl);
     runChoice(5);
+  }
+}
+
+export function hasDisposedGarbage() {
+  return get("_dinseyGarbageDisposed");
+}
+
+export function disposeGarbage() {
+  if (have($item`bag of park garbage`) && !hasDisposedGarbage()) {
+    visitUrl(maintUrl);
+    runChoice(6);
   }
 }
--- a/src/resources/2015/Dinsey.ts
+++ b/src/resources/2015/Dinsey.ts
@@ -1,0 +1,215 @@
+import {
+  availableChoiceOptions,
+  myAdventures,
+  myAscensions,
+  runChoice,
+  visitUrl,
+} from "kolmafia";
+import { getRemainingLiver, have } from "../..";
+import { get, getString, getNumber } from "../../property";
+import { $location, $item } from "../../template-string";
+
+const kioskUrl = "place.php?whichplace=airport_stench&action=airport3_kiosk";
+const maintUrl = "place.php?whichplace=airport_stench&action=airport3_tunnels";
+
+export const questPreferenceKioskText = new Map<string, string>([
+  ["questEStSocialJusticeII", "Racism Reduction"],
+  ["questEStSocialJusticeI", "Sexism Reduction"],
+  ["questEStFishTrash", "Waterway Debris Removal"],
+  ["questEStZippityDooDah", "Compulsory Fun"],
+  ["questEStSuperLuber", "Track Maintenance"],
+  ["questEStGiveMeFuel", "Electrical Maintenance"],
+  ["questEStNastyBears", "Bear Removal"],
+  ["questEStWorkWithFood", "Guest Sustenance Assurance"],
+]);
+
+export const questPreferenceLocation = new Map<string, Location>([
+  [
+    "questEStSocialJusticeII",
+    $location`Uncle Gator's Country Fun-Time Liquid Waste Sluice`,
+  ],
+  ["questEStSocialJusticeI", $location`Pirates of the Garbage Barges`],
+  ["questEStFishTrash", $location`Pirates of the Garbage Barges`],
+  ["questEStZippityDooDah", $location`The Toxic Teacups`],
+  ["questEStSuperLuber", $location`Barf Mountain`],
+  ["questEStGiveMeFuel", $location`The Toxic Teacups`],
+  [
+    "questEStNastyBears",
+    $location`Uncle Gator's Country Fun-Time Liquid Waste Sluice`,
+  ],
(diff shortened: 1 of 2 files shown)


src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,215 @@
...
+      trackingPref: "",
+      completeValue: 0,
+    },
+  ],
+  [
+    "questEStGiveMeFuel",
+    {
+      item: $item`toxic globule`,
+      quantity: 20,
+      trackingPref: "",
```
horrible-little-slime, line 101: I would make this null probably, but I can see why that would be a pain to deal with.

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+export function disposeGarbage() {
+  if (have($item`bag of park garbage`) && !get("_dinseyGarbageDisposed")) {
+    visitUrl(maintUrl);
+    runChoice(6);
+  }
+}
+
+export function fightDinsey() {
```
horrible-little-slime, line 210: I would probably add a macro input to this so you don't have to wrap it with a withMacro

src/resources/index.ts
```diff
@@ -16,6 +16,7 @@ import * as SpookyPutty from "./2009/SpookyPutty";
 import * as TunnelOfLove from "./2017/TunnelOfLove";
 import * as WinterGarden from "./2014/WinterGarden";
 import * as Witchess from "./2016/Witchess";
+import * as Dinsey from "./2015/Dinsey";
```
horrible-little-slime, line 19: I think we try to sort these alphabetically

src/resources/2015/Dinsey.ts
pstalcup: I switched the interface to allow trackingPref and completeValue to be optional (which uses undefined instead of null, but the same idea)

src/resources/2015/Dinsey.ts
pstalcup: I did not do this because how to submit combat is not defined in Libram, so I let the user choose how to submit combat

horrible-little-slime (approved):
I'd still like to add a macro input to this, but I feel like there's probably a reason libram doesn't currently have withMacro, so this is fine.

src/resources/2015/Dinsey.ts
horrible-little-slime: One final remark: I would probably call this `fightWart` on the basis that `Dinsey.fightDinsey()` feels redundant

horrible-little-slime: oh you renamed it
ignore me!

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+  ["questEStSuperLuber", $location`Barf Mountain`],
+  ["questEStGiveMeFuel", $location`The Toxic Teacups`],
+  [
+    "questEStNastyBears",
+    $location`Uncle Gator's Country Fun-Time Liquid Waste Sluice`,
+  ],
+  ["questEStWorkWithFood", $location`Barf Mountain`],
+]);
+
+interface questRequirement {
```
gausie, line 43: This should be TitleCase

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+  item: Item;
```
gausie, line 44: I'd rather use `Item | null` here tbh

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export function acceptedQuest(): string | null {
+  const result = Array.from(questPreferenceKioskText.values()).find(
+    (questPref) => getString(questPref) !== "unstarted"
+  );
+  if (result) {
+    return result;
+  } else {
+    return null;
+  }
```
gausie, line 139: Would this work?

```suggestion
  return Array.from(questPreferenceKioskText.values()).find(
    (questPref) => getString(questPref) !== "unstarted"
  ) || null;
```

horrible-little-slime: I feel like rather than getString I'd prefer get<string>, getProperty, or questStep, but that could be wrong of me.

horrible-little-slime: or, wait, make them StringProperties rather than Strings and use plain old get!

src/resources/2015/Dinsey.ts
pstalcup: For the other two optional parameters, do you prefer string | null, or having them be optional?

gausie: `undefined` is fine for those. In general I *try* to avoid `$proxy[none]` in libram because its not as JSey as using `null` instead. We could even consider getting mafia to return `null` instead of them!

gausie: The issue there would be casting it to polymorphic functions on the other end, though I don't know any functions that would take a `$proxy[none]` and do anything meaningful anyway.

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+    "questEStNastyBears",
+    $location`Uncle Gator's Country Fun-Time Liquid Waste Sluice`,
+  ],
+  ["questEStWorkWithFood", $location`Barf Mountain`],
+]);
+
+interface QuestRequirement {
+  item: Item | null;
+  quantity: number;
+  trackingPref?: string;
```
horrible-little-slime, line 46: make this a `StringProperty`

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+  [$item`keycard γ`, $location`The Toxic Teacups`],
+  [
+    $item`keycard δ`,
+    $location`Uncle Gator's Country Fun-Time Liquid Waste Sluice`,
+  ],
+]);
+
+export function acceptedQuest(): string | null {
+  const result = Array.from(questPreferenceKioskText.values()).find(
+    (questPref) => getString(questPref) !== "unstarted"
```
horrible-little-slime, line 133: make this regular ol' `get`

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+}
+
+export function availableQuests(): string[] {
+  const kioskText = visitUrl(kioskUrl);
+  return Array.from(questPreferenceKioskText)
+    .filter((questPair) => kioskText.includes(questPair[1]))
+    .map((questPair) => questPair[0]);
+}
+
+export function startQuest(questPref: string) {
```
horrible-little-slime, line 154: make this a StringProperty OR make a type consisting specifically of the Dinsey quest props

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+    !foughtWartDinseyThisLife() &&
+    getRemainingLiver() >= 0 &&
+    myAdventures() > 0
+  );
+}
+
+export function fightWartDinsey() {
+  if (canFightWartDinsey()) {
+    visitUrl(maintUrl);
+    runChoice(5);
```
horrible-little-slime, line 208: Is this tested? I don't think you need a `runCombat`, but worth checking.

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+}
+
+export function hasDisposedGarbage() {
```
horrible-little-slime, line 212:
```suggestion
export function hasDisposedGarbage(): boolean {
```

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export function coasterNextTurn() {
+  return get("dinseyRollercoasterNext");
+}
+
+export function foughtWartDinseyThisLife() {
+  return get("lastWartDinseyDefeated") === myAscensions();
+}
+
+export function canFightWartDinsey() {
```
horrible-little-slime, line 196:
```suggestion
export function canFightWartDinsey(): boolean{
```

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+  return (
+    Array.from(keyCardsLocations.keys()).every((keycard) => have(keycard)) &&
+    !foughtWartDinseyThisLife() &&
+    getRemainingLiver() >= 0 &&
+    myAdventures() > 0
+  );
+}
+
+export function fightWartDinsey() {
```
horrible-little-slime, line 205:
```suggestion
export function fightWartDinsey(): void {
```

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+    ) {
+      visitUrl(kioskUrl);
+      if (3 in availableChoiceOptions()) {
+        runChoice(3);
+      }
+    }
+  }
+}
+
+export function coasterNextTurn() {
```
horrible-little-slime, line 188:
```suggestion
export function nextCoasterTurn(): number {
```

`coasterNextTurn` sounds like a boolean

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+        kioskText.indexOf(questPairA[1]) - kioskText.indexOf(questPairB[1])
+    )
+    .map((questPair) => questPair[0])
+    .indexOf(questPref);
+  if (questChoice > -1) {
+    runChoice(questChoice + 1);
+  }
+}
+
+export function finishQuest() {
```
horrible-little-slime, line 169: make this a boolean-returning function that tells us whether you successfully returned

src/resources/2015/Dinsey.ts
```diff
@@ -0,0 +1,221 @@
...
+    (questPref) => getString(questPref) !== "unstarted"
+  );
+  if (result) {
+    return result;
+  } else {
+    return null;
+  }
+}
+
+export function questFinished() {
```
horrible-little-slime, line 142:
```suggestion
export function questFinished(): boolean {
```

gausie:
@pstalcup whats going on here?

gausie:
This is a billion years old and @Tokoeka seems to have something close to done. Will close.

PR #176, opened 2021-12-15
implement rudimentary support for orb predictions

horrible-little-slime:
1. This is reliant on/contains #175
2. I would love for an Orb Understander to double check this--I would be willing to bet my logic is flat-out wrong


gausie:
Doesn't this need to also be exported?

gausie (approved):

gausie:
Better in than out!

PR #177, opened 2021-12-15
Add an additional option to knapsack diet planning:

pstalcup:
* configure which mayo you use per MenuItem

### DIFF
--- a/src/diet/index.ts
+++ b/src/diet/index.ts
@@ -19,11 +19,11 @@ import { have } from "../lib";
 import { get as getModifier } from "../modifier";
 import { get } from "../property";
 import { $effect, $item, $items, $skill, $stat } from "../template-string";
-import { sum } from "../utils";
+import { sum, sumNumbers } from "../utils";
 import { Mayo, installed as mayoInstalled } from "../resources/2015/MayoClinic";
 
-export type DietItem<T> = [MenuItem<T>[], number];
-export type Diet<T> = DietItem<T>[];
+type RawDietEntry<T> = [MenuItem<T>[], number];
+type RawDiet<T> = RawDietEntry<T>[];
 
 type ConsumptionModifiers = {
   forkMug: boolean;
@@ -302,7 +302,7 @@ class DietPlanner<T> {
   consumptionHelpersAndValue(
     menuItem: MenuItem<T>,
     overrideModifiers: Partial<ConsumptionModifiers>
-  ): DietItem<T> {
+  ): RawDietEntry<T> {
     const helpers = [];
     if (
       this.seasoning &&
@@ -377,7 +377,7 @@ class DietPlanner<T> {
     organ: Organ,
     capacity: number,
     overrideModifiers: Partial<ConsumptionModifiers> = {}
-  ): [number, Diet<T>] {
+  ): [number, RawDiet<T>] {
     const submenu = this.menu.filter(
       (menuItem) =>
         menuItem.organ === organ && myLevel() >= menuItem.item.levelreq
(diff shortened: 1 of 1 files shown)


src/diet/index.ts
```diff
@@ -624,3 +633,52 @@ export function planDiet(
     return planFoodBooze;
   }
 }
+
+export function dietEstimatedTurns(diet: [MenuItem[], number][]) {
+  const refinedPalate = diet.some((itemCount) =>
```
gausie, line 638: This looks wrong

src/diet/index.ts
```diff
@@ -624,3 +633,52 @@ export function planDiet(
...
+    } else {
+      const items = menuItems.map((m) => m.item);
+
+      return (
+        sum +
+        count *
+          expectedAdventures(menuItems[menuItems.length - 1].item, {
+            forkMug:
+              items.includes($item`Frosty's frosty mug`) ||
+              items.includes($item`Frosty's frosty mug`),
```
gausie, line 668: As does this

src/diet/index.ts
```diff
@@ -624,3 +633,52 @@ export function planDiet(
...
+
+            seasoning: items.includes($item`Special Seasoning`),
+            mayoflex: items.includes(Mayo.flex),
+            refinedPalate: refinedPalate,
```
gausie, line 672: these can just be

```
refinedPalate,
garish,
```

src/diet/index.ts
pstalcup: got it, will fix

src/diet/index.ts
```diff
@@ -238,10 +242,14 @@ class DietPlanner {
     this.seasoning = menu.find(
       (item) => item.item === $item`Special Seasoning`
     );
-    this.mayoflex =
-      getWorkshed() === $item`portable Mayo Clinic`
-        ? menu.find((item) => item.item === $item`Mayoflex`)
-        : undefined;
+    this.mayoLookup = new Map<Item, MenuItem>();
+    if (mayoInstalled()) {
+      [Mayo.flex, Mayo.zapine].forEach((mayo) => {
```
gausie, line 247: Minor but you're really doing a `find` here not a `forEach` here...

src/diet/index.ts
```diff
@@ -21,6 +20,10 @@ import { get as getModifier } from "../modifier";
 import { get } from "../property";
 import { $effect, $item, $items, $skill, $stat } from "../template-string";
 import { sum } from "../utils";
+import { Mayo, installed as mayoInstalled } from "../resources/2015/MayoClinic";
+
+export type DietItem<T> = [MenuItem<T>[], number];
+export type Diet<T> = DietItem<T>[];
```
phulin, line 26: I think this should be a class with the methods at the bottom on it.

src/diet/index.ts
```diff
@@ -21,6 +20,10 @@ import { get as getModifier } from "../modifier";
 import { get } from "../property";
 import { $effect, $item, $items, $skill, $stat } from "../template-string";
 import { sum } from "../utils";
+import { Mayo, installed as mayoInstalled } from "../resources/2015/MayoClinic";
+
+export type DietItem<T> = [MenuItem<T>[], number];
```
phulin, line 25: I think this should be a class, so we can separate the helpers and main item instead of having them all in one array. I think garbo is our only consumer so it shouldn't be a problem that it's a breaking change.

src/diet/index.ts
```diff
@@ -624,3 +642,144 @@ export function planDiet(
...
+        }
+      })
+    );
+  }
+
+  expectedValue(mpa: number, method: "gross" | "net" = "gross"): number {
+    const adventures = this.expectedAdventures();
+    return (
+      adventures * mpa +
+      this.entries.reduce((dietTotal, dietEntry) => {
```
phulin, line 735: We have a sum function for this

src/diet/index.ts
phulin: I think this is more of a for but we tend to use for... of instead of forEach.

phulin (approved):

PR #178, opened 2021-12-17
Better support for free runs, add free kills

Rinn:
- Move Free Runs into /Resources
- Add Free Kills
- Use shared ActionSource class for both runs and kills sources, which includes potential count (will currently overcount if counters are shared) and cost to use
- Support more robust filtering of actions
- Support merging actions
- Special case `Asdon Martin: Spring-Loaded Front Bumper` to only be used if a backup free run source is found
- Add support for various other free runs and kills


src/resources/FreeKill.ts
```diff
@@ -0,0 +1,167 @@
...
+      equipmentRequirements: () =>
+        new Requirement([], {
+          forceEquip: $items`Lil' Doctor™ bag`,
+        }),
+    }
+  ),
+
+  new ActionSource(
+    $skill`Asdon Martin: Missile Launcher`,
+    () => (!get("_missileLauncherUsed") && AsdonMartin.have() ? 1 : 0),
```
horrible-little-slime, line 75: I think we want .installed(), not .have(); if not, we should account for whether your workshed was changed today. This likely applies to the bumperbanish too,

horrible-little-slime (commented):
I would love to see jsdocs for pretty much all of these functions; I know it's a pain, but you're cooking up a whole new system here and clarity is important.

src/resources/action.ts
```diff
@@ -0,0 +1,141 @@
...
+    preparation: () => {
+      let success = true;
+      options.forEach(
+        (option) =>
+          (success = success && (!option?.preparation || option?.preparation()))
+      );
+      return success;
+    },
+    // Return thet firs familiar if multiples need them
+    familiar: options
```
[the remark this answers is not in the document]
Rinn: yeah idk it's not ideal, maybe i should just make it throw the exception if the merge is impossible

src/resources/FreeRun.ts
```diff
@@ -0,0 +1,272 @@
...
+    if (bumperIndex === -1) return 1;
+    return myTurncount() - parseInt(banishes[bumperIndex + 1]) > 30 ? 1 : 0;
+  },
+  Macro.skill($skill`Asdon Martin: Spring-Loaded Front Bumper`),
+  {
+    preparation: () => AsdonMartin.fillTo(50),
+  }
+);
+
+const freeRunSources: ActionSource[] = [
```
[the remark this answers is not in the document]
Rinn: my thought was to have a separate file for banishers entirely

horrible-little-slime:
Would having it return null be too tedious to code around?

src/resources/FreeRun.ts
```diff
@@ -0,0 +1,272 @@
...
+  // Free limited sources
+  new ActionSource(
+    $familiar`Frumious Bandersnatch`,
+    () =>
+      have($familiar`Frumious Bandersnatch`) &&
+      (have($effect`Ode to Booze`) || getSongCount() < getSongLimit()) &&
+      Bandersnatch.getRemainingRunaways() > 0
+        ? 0
```
Rinn, line 65: whoops didn't implement bander or boots potential yet

src/resources/FreeRun.ts
```diff
@@ -0,0 +1,272 @@
...
+          cliExecute(`fold stinky cheese eye`);
+        return have($item`stinky cheese eye`);
+      },
+    }
+  ),
+
+  new ActionSource(
+    $item`navel ring of navel gazing`,
+    () =>
+      have($item`navel ring of navel gazing`) ? 3 - get("_navelRunaways") : 0,
```
phulin, line 194:
```suggestion
      have($item`navel ring of navel gazing`) ? Math.max(0, 3 - get("_navelRunaways")) : 0,
```

src/resources/action.ts
```diff
@@ -0,0 +1,141 @@
...
+      ]),
+    preparation: () => {
+      let success = true;
+      options.forEach(
+        (option) =>
+          (success = success && (!option?.preparation || option?.preparation()))
+      );
+      return success;
+    },
+    // Return thet firs familiar if multiples need them
```
phulin, line 38:
```suggestion
    // Return the first familiar if multiples need them
```

src/resources/action.ts
```diff
@@ -0,0 +1,141 @@
...
+    this.macro = macro;
+    this.options = options;
+  }
+
+  name(): string {
+    return `${this.source}`;
+  }
+
+  available(): boolean {
+    return this.potential() !== 0;
```
phulin, line 72:
```suggestion
    return this.potential() > 0;
```

This way we don't have to be as careful about making sure things don't go negative (and more robust to tracking bugs, I think).

src/resources/FreeRun.ts
phulin: this is probably the trickiest part conceptually as I'm sure you know, unless you want to maximize for familiar weight every turn. the way I've done it in the past is to have a daily pref for when we've tried and failed, but I don't know if that is generic enough here. Maybe it is.

src/resources/FreeRun.ts
```diff
@@ -0,0 +1,272 @@
...
+    {
+      equipmentRequirements: () =>
+        new Requirement([], { forceEquip: $items`navel ring of navel gazing` }),
+    }
+  ),
+
+  new ActionSource(
+    $item`Greatest American Pants`,
+    () =>
+      have($item`Greatest American Pants`) ? 3 - get("_navelRunaways") : 0,
```
phulin, line 205: same as above

src/resources/FreeRun.ts
```diff
@@ -0,0 +1,272 @@
...
+    Macro.skill($skill`Show your boring familiar pictures`),
+    {
+      equipmentRequirements: () =>
+        new Requirement([], { forceEquip: $items`familiar scrapbook` }),
+    }
+  ),
+
+  new ActionSource(
+    $item`peppermint parasol`,
+    () => 3 - get("_navelRunaways"),
```
phulin, line 231: same as above

src/resources/FreeRun.ts
```diff
@@ -0,0 +1,272 @@
...
+    Macro.item($item`peppermint parasol`),
+    {
+      preparation: () => retrieveItem($item`peppermint parasol`),
+      cost: () => (mallPrice($item`peppermint sprout`) * 5) / 10, // Breaks after 10 uses
```
phulin, line 235:
```suggestion
      cost: () => Math.min(mallPrice($item`peppermint sprout`) * 5, mallPrice($item`peppermint parasol`))) / 10, // Breaks after 10 uses
```

src/resources/FreeRun.ts
```diff
@@ -0,0 +1,272 @@
...
+  ),
+
+  // Expensive unlimited sources
+  ...$items`Louder Than Bomb, divine champagne popper, tennis ball`.map(
+    (item) =>
+      new ActionSource(item, () => -1, Macro.item(item), {
+        preparation: () => retrieveItem(item),
+        cost: () => mallPrice(item),
+      })
+  ),
```
phulin, line 246: Add human musk?

src/resources/action.ts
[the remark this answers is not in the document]
phulin: I think it should just return null if the merge is impossible

src/resources/action.ts
```diff
@@ -0,0 +1,141 @@
...
+    equipmentRequirements: () =>
+      Requirement.merge([
+        ...options.map(
+          (option) =>
+            option.equipmentRequirements?.() ?? new Requirement([], {})
+        ),
+      ]),
+    preparation: () => {
+      let success = true;
+      options.forEach(
```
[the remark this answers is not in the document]
phulin: notionally this should be a for loop since it's stateful - I try not to use `forEach` since I don't like the syntax, can you change to a for loop?

phulin (commented):
Wow, this is really cool! I left some minor comments but can we also move the files to a new directory? maybe `/actions`?

src/resources/action.ts
```diff
@@ -0,0 +1,141 @@
...
+    this.potential = potential;
+    this.macro = macro;
+    this.options = options;
+  }
+
+  name(): string {
+    return `${this.source}`;
+  }
+
+  available(): boolean {
```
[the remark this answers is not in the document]
Rinn: Yeah. I suppose we could do a speculative maximize the first time and use that to estimate the max bander\boots runs?

PR #184, opened 2021-12-26
Fix maximize cache issues.

phulin:
* Fix compatibility with preventSlot - this was the cause of "failure to equip none" messages.
* Make sure familiar equip is in the familiar slot. This could cause problems with lefty.


phulin:
Merging this so I can put out a release

PR #185, opened 2021-12-26
Return success/failure on maximize commands.

phulin:


src/maximize.ts
```diff
@@ -468,13 +469,14 @@ export function maximizeCached(
     applyCached(cacheEntry, fullOptions);
     if (verifyCached(cacheEntry)) {
       logger.info(`Equipped cached ${objective}`);
-      return;
+      return true;
```
horrible-little-slime, line 472: Do we assume that the outfit succeeds?

gausie (approved):

src/maximize.ts
phulin: Yeah but we should fix that separately (and anyway we verify after)

PR #187, opened 2021-12-27
Null on error when converting a property from string to another type.

pstalcup:
Add "validate" method for proxy class types to verify their
string representation matches what is in the property

### DIFF
--- a/src/property.ts
+++ b/src/property.ts
@@ -13,7 +13,15 @@ import {
   PropertyValue,
 } from "./propertyTyping";
 
-import { NumericOrStringProperty, NumericProperty } from "./propertyTypes";
+import {
+  FamiliarProperty,
+  LocationProperty,
+  MonsterProperty,
+  NumericOrStringProperty,
+  NumericProperty,
+  PhylumProperty,
+  StatProperty,
+} from "./propertyTypes";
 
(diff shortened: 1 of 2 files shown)


src/property.ts
```diff
@@ -43,13 +51,24 @@ type MafiaClasses =
...
   Type: typeof MafiaClass & (new () => T)
 ): ((property: string, default_?: T | null) => T | null) =>
   createPropertyGetter((value) => {
     if (value === "") return null;
-    const v = Type.get<T>(value);
-    return v === Type.get<T>("none") ? null : v;
+    try {
+      const v = Type.get<T>(value);
+      return v === Type.get<T>("none") ? null : v;
+    } catch {
```
horrible-little-slime, line 69: Will this lead to a Mafia briefly flashing red every time we get an error? I'm fine with that, of course, just curious

horrible-little-slime (commented):
This looks great! I am not comfortable approving it myself, but this reads like it does exactly what I want

src/property.ts
gausie: What exception is thrown here? Can you narrow the catch expression?

horrible-little-slime:
Resolved by #217

PR #195, opened 2022-01-02
Add ActionSource concept and include free runs + free kills.

phulin:
This is work on top of Rinn's #178 - I've added JSDoc for all exported members and made a few other changes. Only functional change was to make the finder only include expensive unlimited sources (e.g. LTBs) if a maximum cost is specified.

Rinn's original description:

*Move Free Runs into /Resources
*Add Free Kills
*Use shared ActionSource class for both runs and kills sources, which includes potential count (will currently overcount if counters are shared) and cost to use
*Support more robust filtering of actions
*Support merging actions
*Special case Asdon Martin: Spring-Loaded Front Bumper to only be used if a backup free run source is found
*Add support for various other free runs and kills


src/actions/FreeRun.ts
```diff
@@ -0,0 +1,300 @@
...
+
+const freeRunSources: ActionSource[] = [
+  // Free limited sources
+  new ActionSource(
+    $familiar`Frumious Bandersnatch`,
+    () =>
+      have($familiar`Frumious Bandersnatch`) &&
+      (have($effect`Ode to Booze`) || getSongCount() < getSongLimit()) &&
+      Bandersnatch.getRemainingRunaways() > 0
+        ? 0
```
Rinn, line 59: Right, I never had this return the potential, so it should for now just return Bandersnatch.getRemainingRunaways(). Same for boots.

Rinn (approved):
Yeah otherwise I think this is pretty good, thanks for running with this.

PR #201, opened 2022-01-07
Asdon tweaks

lewismd13:


horrible-little-slime (approved):
I'm going to go ahead and assume that canInteract does what I assume it does

lewismd13:
> I'm going to go ahead and assume that canInteract does what I assume it does

yeah it's a weird function name for "are you in casual/aftercore/broke ronin"

PR #208, opened 2022-01-12
Shavehat tracking

horrible-little-slime:


src/resources/2021/DaylightShavings.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const cycle = buffCycle();
+
+  const index = cycle.indexOf(currentBuff);
+
+  const newIndex = (1 + index) % 11;
+  return cycle[newIndex];
+}
+
+/**
```
docrostov, line 53: i think all these functions make sense but buffsUntil is a little unintuitive, though i see why it is useful after our convo on discord. i would recommend a different docstring. here is my attempt to punch it up. 
```/**
 * Returns the number of shaving hat activations we expect it'll take to get to a given buff, given our knowledge about the user's current buffCycle. Returns 1 for the next buff, 2 for the one after that, and so on. Returns 11 for the most recent buff. This is useful for both ascertaining whether or not to buy SGEEAs to cycle to your next buff and timing out how long you would need to wait without SGEEA interaction.
 * @param buff The shaving buff in question (as an $effect)
 * @returns The number of shaving hat activations necessary to get the shaving buff in @param; null if said buff is not granted by the shaving hat.
 */```

File: src/resources/2021/DaylightShavings.ts [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+
+export function hasBuff(): boolean {
+  return buffs.some((buff) => haveItem(buff));
+}
+/**
+ * Calculates and returns the cycle of buffs that the hat should cycle through.
+ * @param playerclass The class to generate a cycle for
+ * @returns An ordered array consisting of the cycle for this class. The first element of the array will be the first buff a player should expect to get in a given ascension.
+ */
+export function buffCycle(playerclass = myClass()): Effect[] {
```
docrostov, line 26: my assumption is the answer to this is "i dunno, sue me" but do we know what happens to shaving helmet cycle when you drop path? that is the one place i could see this breaking, and while it doesn't seem like a major vector for bugs, it is probably worth at least doing minor digging to see if this is a possible breakpoint for libram, or putting a catch here to output a null cycle the day you have dropped path or ascended out of an avatar.

horrible-little-slime: If someone wants to spade this, I will happily update the code. I've made sure it no longer explodes if you somehow submit astral spirit, but I don't think we need to independently check this out before publishing.

PR #210, opened 2022-01-12
Tweaks to ascension code.

phulin:
Fix one bug, clean up some stuff.


src/ascend.ts
```diff
@@ -21,7 +20,29 @@ export enum Lifestyle {
   hardcore = 3,
 }
 
-function toMoonId(moon: string | number, playerClass: Class): number {
+type MoonSign =
```
horrible-little-slime, line 23: I almost want to break this into like 10 types and make it a union of those types, even though there's no utility to that.

horrible-little-slime (approved):

src/ascend.ts
phulin: at some point we should have a moon sign type either in mafia or in libram that can be used anywhere.

PR #213, opened 2022-01-14
Get effect from The Mad Tea Party

Rinn:


src/resources/2010/LookingGlass.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @param effect The effect to check
+ * @returns The number of characters required to get an effect
+ */
+export function findTeaPartyHatLength(effect: Effect): number {
+  const effectId = toInt(effect);
+  if (
+    effectId < toInt(firstTeaPartyEffect) ||
+    effectId > toInt(lastTeaPartyEffect)
+  ) {
+    throw `Invalid Mad Tea Party effect ${effect}`;
```
phulin, line 37: libram style has throws only on Error objects, not raw strings

src/resources/2010/LookingGlass.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @returns If any valid hat is in inventory
+ */
+export function haveTeaPartyHat(effect: Effect): boolean {
+  const characters = findTeaPartyHatLength(effect);
+  if (!characters) {
+    throw `Invalid Mad Tea Party effect ${effect}`;
+  }
+  return (
+    Item.all().find(
+      (item) => validTeaPartyHat(item, characters) && have(item)
```
phulin, line 67: you need to be able to equip it too, not just have it

Rinn: It's checked in validTeaPartyHat:
```/**
 * Checks if an item is a valid hat for The Mad Tea Party
 * @param item The item to check
 * @param characters The desired nameLength for The Mad Tea Party effect
 * @returns If this item is a valid hat for The Mad Tea Party
 */
function validTeaPartyHat(item: Item, characters: number): boolean {
  if (toSlot(item) !== $slot`hat` || !canEquip(item)) {
    return false;
  } else if (characters <= 5) {
    return item.nameLength <= 5;
  } else if (characters >= 30) {
    return item.nameLength >= 30;
  }
  return characters === item.nameLength;
}
```

phulin: thanks, sorry

phulin (approved):

PR #214, opened 2022-01-14
Eudora-fix

horrible-little-slime:


Rinn (approved):

Rinn (changes requested):
Actually maybe let's go with the enum stuff we talked about in chat.

Rinn (approved):

PR #217, opened 2022-01-17
Overhaul property system for speed.

phulin:
While really, really nice, our well-typed property system is very nonperformant while doing live development in VS Code or other IDEs. These edits maintain near-full fidelity to the original code; we only lose type checking accuracy on multiple-property functions like `withProperties`.


phulin:
API question: should `getBoolean` return null if the property doesn't exist? Or should there be a default default of false?

phulin:
I reverted back to the existing behavior so as not to cause like 500 errors in garbo.

pstalcup (approved):

phulin:
Requesting your review sam whenever you have time + energy + inclination, no rush.

horrible-little-slime:
It's worth mentioning that this devours/fulfills #187

PR #218, opened 2022-01-19
Add "Session" for tracking mafia session results

pstalcup:


src/session.ts
```diff
@@ -0,0 +1,272 @@
...
+}
+
+/**
+ * A wrapper around tracking items and meat gained from this session
+ * Smartly handles foldables being added/removed based on their state
+ * Provides operations to add sessions and subtract Sessions so you can isolate the value of each Session using a baseline
+ * @member meat the raw meat associated with this Session
+ * @member items a map representing the items gained/lost during this Session
+ */
+export class Session {
```
horrible-little-slime, line 140:
```suggestion
export default class Session {
```

src/session.ts
```diff
@@ -0,0 +1,272 @@
...
+import { getFoldGroup } from "./lib";
+import { $item, $items } from "./template-string";
+import { sumNumbers } from "./utils";
+
+/**
+ * Return a mapping of the session items, mapping foldable items to a single of their forms
+ * @returns the item session results, with foldables mapped to a single of their folding forms
+ */
+function mySessionItemsWrapper(): Map<Item, number> {
+  const manyToOne = (primary: Item, mapped: Item[]): [Item, Item][] =>
```
horrible-little-slime, line 17: I kinda of want to yoink these `const`s out--it feels like they could be applicable outside of this function

src/session.ts
```diff
@@ -0,0 +1,272 @@
...
+    manyToOne(item, getFoldGroup(item));
+
+  const itemMappings = new Map<Item, Item>([
+    ...foldable($item`liar's pants`),
+    ...foldable($item`ice pick`),
+    ...manyToOne($item`Spooky Putty sheet`, [
+      $item`Spooky Putty monster`,
+      ...getFoldGroup($item`Spooky Putty sheet`),
+    ]),
+    ...foldable($item`stinky cheese sword`),
```
horrible-little-slime, line 30: Is there a better way to handle this other than listing everything?

src/session.ts
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * A wrapper around tracking items and meat gained from this session
+ * Smartly handles foldables being added/removed based on their state
+ * Provides operations to add sessions and subtract Sessions so you can isolate the value of each Session using a baseline
+ * @member meat the raw meat associated with this Session
+ * @member items a map representing the items gained/lost during this Session
+ */
+export class Session {
+  meat: number;
```
horrible-little-slime, line 141: feels like maybe these should be private with getters?

horrible-little-slime (commented):
Looks good to me!

src/session.ts
phulin: i don't think we are using default exports anywhere else in libram

src/session.ts
phulin: nah, i think this is fine

phulin (approved):

src/session.ts
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  /**
+   * Combine the contents of sessions
+   * @param sessions the set of sessions to combine together
+   * @returns a new session representing the difference between a and b
+   */
+  static add(...sessions: Session[]): Session {
+    return sessions.reduce((previoussession, currentsession) =>
+      previoussession.add(currentsession)
```
phulin, line 237:
```suggestion
    return sessions.reduce((previousSession, currentSession) =>
      previoussSssion.add(currentSession)
```

src/session.ts
horrible-little-slime: When I first wrote `Dungeon.ts`, gausie suggested that I export `Dungeon` as default, that's what I was basing this on

PR #227, opened 2022-01-26
Compile a version of libram for require().

phulin:
This will enable `require("libram")` from the Mafia command line. You could even make the following alias:

```ts
alias jsl => js { Object.assign(globalThis, require("libram")); }
```

and then:

```
> jsl AsdonMartin.installed();

Returned: false

> jsl $item`hair spray`

Returned: hair spray
name => hair spray
...
```


phulin:
Unfortunately, looks like a bug in our value code block this from working somewhere. I'll have to track it down.

phulin:
Never mind - the only issue is that we can't render the entire `libram` object because our JS -> ASH translation code can't translate enums (inconsistent typing of keys and values).

src/Clan.ts
```diff
@@ -29,40 +29,6 @@ export interface Rank {
   id: number;
 }
 
-export class ClanError extends Error {
-  reason?: Error;
```
gausie, line 33: My poor code

phulin: sorry buddy, we weren't using it

babel.config.js
```diff
@@ -0,0 +1,18 @@
...
+module.exports = function (api) {
+  api.cache(true);
+  return {
+    exclude: [],
+    presets: [
+      "@babel/preset-typescript",
+      [
+        "@babel/preset-env",
+        {
+          targets: { rhino: "1.7.13" },
```
gausie, line 12: Do you want to bump this?

phulin:
Yes

package.json
```diff
@@ -22,13 +25,12 @@
     "dist/**/*.d.ts"
   ],
   "devDependencies": {
-    "@babel/cli": "^7.14.8",
     "@babel/compat-data": "^7.15.0",
-    "@babel/core": "^7.15.0",
+    "@babel/core": "^7.16.12",
```
gausie, line 29: Presumably that bump means tha these need to be bumped to 7.17 or whatever

phulin: done

gausie (approved):

PR #229, opened 2022-01-28
Add missing imports for new kolmafia structure.

phulin:
Waiting on kolmafia-js update.


gausie (approved):
What are the resolver changes?

phulin:
That's just for eslint - I don't think the import plugin was configured correctly, so it wasn't detecting the missing symbols here.

PR #232, opened 2022-01-28
Mummery

horrible-little-slime:
Regexps are tested

`Meat Drop: [30*fam(Robortender)],Item Drop: [25*fam(Trick-or-Treating Tot)],MP Regen Min: [4*fam(Puck Man)], MP Regen Max: [5*fam(Puck Man)],Experience (Moxie): [2*fam(Mini-Hipster)],Experience (Muscle): [3*fam(Artistic Goth Kid)],Experience (Mysticality): [4*fam(Melodramedary)],HP Regen Min: [8*fam(Levitating Potato)], HP Regen Max: [10*fam(Levitating Potato)],` is a fully maxed out mumming trunk example; my one concern here is that for hp and mp regen, we actually get two modifiers per thing rather than 1, but I dont' want it to be a [NumericModifier, number][]` or anything like that, so 🤷 

I plan on eventually adding support for determining which familiars get bonuses from mummery costumes, and for putting on costumes, but today is not that day. This primarily exists for the change it applies to CS


src/resources/2017/MummingTrunk.ts
```diff
@@ -0,0 +1,32 @@
+import { Familiar, toFamiliar } from "kolmafia";
+import { NumericModifier } from "../../modifierTypes";
+import { get } from "../../property";
+
+function toModifier(input: string): NumericModifier {
+  const regExp = new RegExp(/(Experience)+ \(([^)]+)\)/);
```
gausie, line 6: Why do you want to match `ExperienceExperienceExperience (Moxie)`? Also, just do `Experience \((.*?)\)`

src/resources/2017/MummingTrunk.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const matcher = input.match(regExp);
+  if (matcher) {
```
gausie, line 9:
```ts
return (matcher ? `${matcher[2]} Experience` : input) as NumericModifier;
```

src/resources/2017/MummingTrunk.ts
```diff
@@ -0,0 +1,32 @@
...
+    return `${matcher[2]} Experience` as NumericModifier;
+  }
+  return input as NumericModifier;
+}
+
+export function currentCostumes(): Map<Familiar, [NumericModifier, number]> {
```
gausie, line 15: Document this fn please

PR #234, opened 2022-02-03
fix counting of familiars for item test

horrible-little-slime:
Equipment with significant gear, like the tot and disembodied hand, have the enchants of their gear counted as part of the familiar version of numericModifier.

### DIFF
--- a/src/challengePaths/2015/CommunityService.ts
+++ b/src/challengePaths/2015/CommunityService.ts
@@ -29,7 +29,6 @@ import {
   $class,
   $effect,
   $familiar,
-  $familiars,
   $item,
   $items,
   $slot,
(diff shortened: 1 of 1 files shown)


lewismd13 (commented):
I don't think I fully understand using this approach. For example, it will not catch a pet rock with luck incense equipped as 25% item drop. Unless I'm missing something, you can just do numericModifier(equippedItem($slot`familiar`), "Item Drop") and catch all possible equips without messing around with the actual familiar's item drop effect that doesn't count, right?

lewismd13 (approved):

src/challengePaths/2015/CommunityService.ts
```diff
@@ -312,7 +321,11 @@ export const BoozeDrop = new Test(
     return (
       60 -
       multiplier *
-        Math.floor((getModifier("Item Drop") - familiarItemDrop) / 30 + 0.001) -
+        Math.floor(
+          (getModifier("Item Drop") - familiarItemDrop - familiarBoozeDrop) /
+            30 +
+            0.001
```
gausie, line 327: Can we get a comment above for why we're adding 0.001? Same as below. I know this is not introduced in this PR

PR #238, opened 2022-02-05
create a maxTurns parameter for the Test.run() method; stop submitting errored tests

horrible-little-slime:


src/challengePaths/2015/CommunityService.ts
```diff
@@ -129,20 +129,29 @@ class Test {
    * @param beCertain Whether we should check council.php instead of mafia to determine whether the test is complete.
    * @returns The output of the prepare function given, or null if the test is already complete.
    */
-  run<T>(prepare: () => T, beCertain = false): T | null {
+  run<T>(
+    prepare: () => T,
+    beCertain = false,
+    maxTurns: number | null = null
```
gausie, line 135: if you want to be wanky you can do `maxTurns = Infinity` here and save some logic further down

src/challengePaths/2015/CommunityService.ts
```diff
@@ -129,20 +129,29 @@ class Test {
...
     const startTime = Date.now();
     const startTurns = myTurncount();
 
+    let submitTest = true;
     try {
       return prepare();
+    } catch {
+      submitTest = false;
     } finally {
       const prediction = this.predictor();
```
gausie, line 150: I don't fully understand why we have a finally block here. So in either case we want to log that we did (or didn't do) the test? Why are we logging regular tests the same as failed tests. Why are we returning the value of a prepare function instead of a do function? This is smelly.

src/challengePaths/2015/CommunityService.ts
horrible-little-slime: I do want to be wanky. I was considering this, and now I'll go ahead and pull the trigger.

src/challengePaths/2015/CommunityService.ts
horrible-little-slime: I originally wrapped the whole thing in a try...finally because it was paralleling the various withX functions.

I think you're right that this is smelly, but `this.do()` just submits the test. I think making this return a boolean instead of the generic typing does make more sense, though.

src/challengePaths/2015/CommunityService.ts
```diff
@@ -129,31 +129,37 @@ class Test {
    * @param beCertain Whether we should check council.php instead of mafia to determine whether the test is complete.
```
gausie, line 129: This needs a new line

src/challengePaths/2015/CommunityService.ts
```diff
@@ -129,31 +129,37 @@ class Test {
...
    * @returns The output of the prepare function given, or null if the test is already complete.
    */
-  run<T>(prepare: () => T, beCertain = false): T | null {
+  run(prepare: () => void, beCertain = false, maxTurns = Infinity): boolean {
     const finishedFunction = () =>
```
gausie, line 133: `const finishedFunction = beCertain ? this.verifyIsDone : this.isDone`

src/challengePaths/2015/CommunityService.ts
```diff
@@ -129,31 +129,37 @@ class Test {
...
+    if (prediction < maxTurns) {
       this.do();
+    }
 
-      const loggedTest = {
-        predictedTurns: prediction,
-        turnCost: myTurncount() - startTurns,
-        seconds: (Date.now() - startTime) / 1000,
-      };
+    const loggedTest = {
```
gausie, line 152: This could move into the if block (optional)

gausie (commented):
One optional further change

src/challengePaths/2015/CommunityService.ts
gausie: For `maxTurns` I mean

src/challengePaths/2015/CommunityService.ts
horrible-little-slime: https://github.com/Loathing-Associates-Scripting-Society/libram/pull/231
Doing this unbinds the `verifyIsDone` and `isDone` methods from their `this`, and the current setup is more elegant than `const finishedFunction = (beCertain ? this.verifyIsDone : this.isDone).bind(this)`

gausie (approved):

PR #245, opened 2022-02-11
added resting to chateau mantegna

ktpolanski:


horrible-little-slime (dismissed):

src/resources/2015/ChateauMantegna.ts
```diff
@@ -18,6 +26,17 @@ export function fightPainting(): string {
   return runCombat();
 }
 
+export function rest(): void {
+  visitUrl("place.php?whichplace=chateau&action=chateau_restbox");
+}
+
+export function restIfFree(): boolean {
```
gausie, line 33: This function should probably be taken out into a general library function, and could even take a callback that would dictate what rest it did

ktpolanski:
Once this suggestion was made, I talked the issue over with phred. I'm not a huge fan of a generic `restIfFree()` with a callback as it's not the user friendliest in my opinion. However, it does make the code future-proof, and seems like it can be easily locally used for specific rest types as per phred's recent idea.

gausie:
@ktpolanski sorry for letting this get stale - want to fix the conflicts?

gausie:
Going to consider this dead. Please please feel free to reopen and we'll review quickly

PR #250, opened 2022-02-16
restructure CommunityService.ts

horrible-little-slime:
This does a bunch of stuff!

- Changes the Test class to the CommunityService class, makes the constructor private, exports the class, and makes the 11 instances statics on the class
- Change the return typing of CommunityService.test.run() to `"failed" | "already completed" | "completed"`, which corresponds to pretty much what you'd expect
- Allows the `prepare` function called in `run()` to optionally return a numeric value representing the expected number of turns spent during the preparation process (the big shout here is Simmer for spell damage)
- Fiddles around a little bit with documentation and syntax


horrible-little-slime:
I have been running a `phccs` build from this without bugs for ~3 days

ktpolanski:
I've been using this in `dc6s` for a while now, without hiccups.

burningbman (approved):

PR #251, opened 2022-02-16
use two separate xpaths to find clan Ids and Names. Also, fix Clan.with() issue

horrible-little-slime:
For some reason, the current setup is periodically causing errors for some users. This setup seems not to.


horrible-little-slime:
I feel like I haven't heard anyone have this issue in ages so I'm closing this

PR #255, opened 2022-03-01
add rudimentary retro cape support

horrible-little-slime:


src/resources/2020/RetroCape.ts
```diff
@@ -0,0 +1,124 @@
...
+};
+type Hero = keyof typeof Heroes;
+type Mode = "hold" | "thrill" | "kiss" | "kill";
+
+export const currentHero = () => get("retroCapeSuperhero") as Hero;
+export const currentMode = () => get("retroCapeWashingInstructions") as Mode;
+
+const modeMap = new Map<[Hero, Mode], Modifiers>([
+  [
+    ["vampire", "hold"],
```
burningbman, line 39: Did you consider structing it as:
```
{
  vampire: {
    hold: {},
    thrill: {}
}
```
It's not a huge deal since it isn't exported, but your setup struck me as a bit confusing at first.
I guess it can't be a map that way, but a nested object makes more sense to me for this data.

burningbman (approved):

horrible-little-slime:
I thoughr about that. I don't quite remember why I didn't do it. Maybe I'll
set it up that way in a bit.

PR #256, opened 2022-03-01
Power Plant support

horrible-little-slime:
harvest() is a meme implementation, but it is also beat for beat what gausie requested


burningbman:
Why do you want to use breakfast instead of visitUrl to inventory and choice?

horrible-little-slime:
This is a joke PR for sam's benefit; closing to avoid clogging

PR #263, opened 2022-03-17
Add functionality for actual CS test cost

MrFizzyBubbs:


src/challengePaths/2015/CommunityService.ts
```diff
@@ -155,7 +155,7 @@ export default class CommunityService {
 
     const prediction = this.predictor();
 
-    if (prediction <= maxTurns) {
+    if (beCertain ? this.actualCost() : prediction <= maxTurns) {
```
s-k-z, line 158: It looks like this case needs some parenthesis to specify order of operations.
```
if ((beCertain ? this.actualCost() : prediction) <= maxTurns) {
```

horrible-little-slime (approved):

PR #265, opened 2022-03-25
Check if the player can ascend today

MrFizzyBubbs:


MrFizzyBubbs:
[Kasekopf ](http://www.koldb.com/player.php?name=kasekopf) has 4 runs that all are labelled with 3/20/22 so I guess this isn't feasible

PR #271, opened 2022-04-05
use Infinity as a return value instead of null for Counter.get

horrible-little-slime:
Rever and I had a conversation about this. My position boils down to the following:
- `null` as a value has a big disadvantage of making the return type always be `number | null`, which means any time you want to use a comparison operator, you need to throw a `??` in there
- `NaN`'s biggest single disadvantage is that `NaN === NaN` is false. This means that it's very difficult to have a script actively identify unset counters (could we do `const x = Counter.get("whatever"); x !== x`?). It also returns false for all other comparison operators, which is probably bad behavior. Its biggest advantage, however, is that it's falsy. But so is 0, and 0 is a pretty important value for counters.
- `Infinity` behaves the way I personally want it to with all comparisons. It says "this counter isn't happening yet", and it also says "this counter won't happen for a very long time", and it doesn't say "this counter is overdue". It also allows us to identify unset counters via `Counter.get("whatever") === Infinity`.

I'm open to debate as to whether the return value for unset counters should be `null`, `Infinity`, or `NaN`. But I'm also confident enough that the best answer is `Infinity` that I'm making this PR.

### DIFF
--- a/src/counter.ts
+++ b/src/counter.ts
@@ -5,13 +5,13 @@ import { cliExecute, getCounter, getCounters } from "kolmafia";
  * @param counter The name of the counter in question
  * @returns null if the counter does not exist; otherwise returns the duration of the counter
  */
-export function get(counter: string): number | null {
+export function get(counter: string): number {
   const value = getCounter(counter);
   //getCounter returns -1 for counters that don't exist, but it also returns -1 for counters whose value is -1
   if (value === -1) {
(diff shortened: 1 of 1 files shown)


gausie:
I think this is fine, though I have a few suggestions:

* Make sure to update the documentation - most important step
* I would add an additional boolean library function that checks the existance of a counter, that way someone who just wants to check can use that function instead of having to think about `Infinity`

lewismd13 (approved):

PR #272, opened 2022-04-09
Fix getBanishedMonster

Tokoeka:
Add special cases for Use the Force, and the two skills that have matching item names

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -413,8 +413,15 @@ export function getBanishedMonsters(): Map<Item | Skill, Monster> {
     if (foe === undefined || banisher === undefined) break;
     // toItem doesn"t error if the item doesn"t exist, so we have to use that.
     const banisherItem = toItem(banisher);
-    const banisherObject = [Item.get("none"), null].includes(banisherItem)
-      ? Skill.get(banisher)
+    const banisherObject = [
+      Item.get("none"),
+      Item.get(`training scroll:  Snokebomb`),
+      Item.get(`tomayohawk-style reflex hammer`),
+      null,
+    ].includes(banisherItem)
+      ? banisher.toLowerCase() === "saber force"
+        ? Skill.get(`7311`)
+        : Skill.get(banisher)
       : banisherItem;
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
This ternary is getting a little chonky; could we break this up into a longer codeblock, maybe breaking it off into a separate function if appropriate?

Tokoeka:
> This ternary is getting a little chonky; could we break this up into a longer codeblock, maybe breaking it off into a separate function if appropriate?

Ternary changed out into a selection of if/else statements.

src/lib.ts
```diff
@@ -413,17 +413,20 @@ export function getBanishedMonsters(): Map<Item | Skill, Monster> {
...
+      result.set(Skill.get(`7311`), Monster.get(foe));
+    } else if (
+      [
+        Item.get("none"),
+        Item.get(`training scroll:  Snokebomb`),
+        Item.get(`tomayohawk-style reflex hammer`),
+        null,
+      ].includes(banisherItem)
+    ) {
+      result.set(Skill.get(banisher), Monster.get(foe));
```
horrible-little-slime, line 426: Do we want to check that this isn't $skill[none]? You and I know that it won't happen for snokebomb or reflex hammer, but it could happen for null or $item[none]

src/lib.ts
```diff
@@ -413,17 +413,20 @@ export function getBanishedMonsters(): Map<Item | Skill, Monster> {
...
-      Item.get(`tomayohawk-style reflex hammer`),
-      null,
-    ].includes(banisherItem)
-      ? banisher.toLowerCase() === "saber force"
-        ? Skill.get(`7311`)
-        : Skill.get(banisher)
-      : banisherItem;
-    result.set(banisherObject, Monster.get(foe));
+    if (banisher.toLowerCase() === "saber force") {
+      result.set(Skill.get(`7311`), Monster.get(foe));
```
horrible-little-slime, line 417: Is there a reason we're using X.get() instead of template strings here?

horrible-little-slime (approved):

PR #275, opened 2022-04-11
fine-tune logic used for shrugging songs in Mood.execute

horrible-little-slime:
The goal here is to reduce the amount of unnecessary shrugging we do, but always shrug when able. As such, we:
 - only shrug when you don't have enough "room" for the song we're casting
 - only shrug one song when we shrug
 - Prioritize songs that "compete" for this slot, then songs that are slotless, then songs that "compete" for another slot (and lose)

Very open to vigorous debate as to what optimal behavior looks like here.

### DIFF
--- a/src/mood.ts
+++ b/src/mood.ts
@@ -26,7 +26,13 @@ import {
   use,
   useSkill,
 } from "kolmafia";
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
Obsoleted by other mood changes planned

PR #285, opened 2022-04-28
Grey Goose support

horrible-little-slime:


horrible-little-slime:
@docrostov just putting this back on your radar--just want to know if these are all the utility functions you'd like access to

src/resources/2022/GreyGoose.ts
```diff
@@ -0,0 +1,53 @@
...
+import { get } from "../../property";
+import { $familiar, $item, $skill } from "../../template-string";
+
+export const goose = $familiar`Grey Goose`;
+
+export function have(): boolean {
+  return have_(goose);
+}
+
+export function currentExperience(): number {
```
docrostov, line 12: does the familiar data type's "experience" function default to missing, or zero? according to the wiki for [proxy records](https://wiki.kolmafia.us/index.php/Proxy_Records), it defaults to 0, which to me would indicate this wouldn't actually work. since a zero XP goose doesn't *always* have 81-100 XP incoming, it is possible the user used it up in a fight and didn't conclude the fight naturally (a la saber, freerun, etc). i could be wrong tho! was this tested?

src/resources/2022/GreyGoose.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export function currentWeight(): number {
+  return Math.min(Math.floor(Math.sqrt(currentExperience())), 20);
+}
+
+export function expectedDrones(weight = currentWeight()): number {
+  return Math.max(0, weight - 5);
+}
+
+export function expectedExperience(weight = currentWeight()): number {
```
docrostov, line 26: this needs a grey you switch. if the user is in grey you, the experience skills grant you (weight - 5)^2 fullstats rather than the cube of weight-5 substats. this one probably needs a docstring due to this small complexity.

src/resources/2022/GreyGoose.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export function expectedMeat(weight = currentWeight()): number {
+  return Math.pow(Math.max(weight - 5, 0), 4);
+}
+
+export function hasMeatified(): boolean {
+  return get("_meatifyMatterUsed");
+}
+
+export function fightsUntil(
```
docrostov, line 38: in theory, this is a good and necessary function. in practice, i'm not sure it's strictly usable in this form in the way it seems most likely to be applied, as it is quite subject to change in most scripting contexts depending on things within the script's control. as an example, let's say you are writing a grey you looping script. you run fightsUntil and it says it's 3 turns to get a new goose charge. the script does not want to accidentally encounter an adventure monster, so it goes to do something else, and modifier maximizer puts new things on, changing the fightsUntil. this leads to weird behavior where it just keeps hopping around and never re-optimizing for famxp, unless it is coded around explicitly. just feels like the way this is set up is both perfectly fine for a simple implementation but a potentially annoying one that could easily get stuck in a weird loop with a simple loop script -- this is especially rough if the user uses something like the susie's buff from the fortune teller, where this doesn't actually address the lost value of potentially not fitting as many resorbs as possible into the 100 turns.

i think one way to "fix" this is to (in addition to fightsUntil) have optional returns that evaluate how much familiar XP you'd need to cut a turn off of fightsUntil and how much familiar XP you'd need to lose to add a turn to fightsUntil. that way, those scripts can set up a bounding condition where (say) it forces modmax to keep you at +6 familiar XP however possible to maintain the expected fightsUntil and not get caught in a bad local minimum. it's not perfect but it would at least start to address this issue.

src/resources/2022/GreyGoose.ts
horrible-little-slime: It does, and I just accounted for that by checking the only thing that can make it roll back down to 0 exp

src/resources/2022/GreyGoose.ts
horrible-little-slime: Declaring out of scope of the first PR

docrostov (approved):
switching my review to approve but i will once again note that fightsUntil is... a mite bit questionable. still, good to have this in libram and it looks right to me.

horrible-little-slime:
> switching my review to approve but i will once again note that fightsUntil is... a mite bit questionable. still, good to have this in libram and it looks right to me.

That's future me's problem. Or future you!

PR #286, opened 2022-04-29
create & use AscensionPrepError class, alongside the AscendError class

horrible-little-slime:
I ended up not ever using the little utility function I cooked up as a type guard, but it felt appropriate to use it as is


src/ascend.ts
```diff
@@ -40,9 +51,88 @@ export class AscendError extends Error {
...
+const isWorkshed = createStringUnionTypeGuardFunction(worksheds);
+const isGarden = createStringUnionTypeGuardFunction(gardens);
+const isEudora = createStringUnionTypeGuardFunction(eudorae);
+const isDesk = createStringUnionTypeGuardFunction(desks);
+const isNightstand = createStringUnionTypeGuardFunction(nightstands);
+const isCeiling = createStringUnionTypeGuardFunction(ceilings);
+
+export class AscensionPrepError extends Error {
+  cause: string;
+  constructor(cause: string) {
```
burningbman, line 108: I'd change this to take the current value as well as the cause. Then then the string building logic is easier.

src/utils.ts
```diff
@@ -116,3 +116,14 @@ export function invertMap<T1, T2>(map: Map<T1, T2>): Map<T2, T1> {
   }
   return returnValue;
 }
+
+/**
+ * Creates a Type Guard function for a string union type defined via an array as const.
+ */
+export function createStringUnionTypeGuardFunction<T extends string>(
```
burningbman, line 123: I don't understand what this doing.

horrible-little-slime: We have a lot of string literal types defined by arrays with the `as const` flag, especially in ascend.ts. This creates type guard functions for those types, using the array from which they are defined. 

In this particular PR, we never really use the type guard aspect of it, only the boolean return value. But I figured it's probably helpful to have a type guard function.

PR #287, opened 2022-04-29
florist fixes round iii

horrible-little-slime:
@MrFizzyBubbs 👀


MrFizzyBubbs (changes requested):
These changes work, however, the the planting url is incorrect. It should use "option" instead of "whichoption": https://github.com/Loathing-Associates-Scripting-Society/libram/blob/c589e1d69c2d5a609849afc8e94e98eeda7bd648/src/resources/2013/Florist.ts#L78

MrFizzyBubbs:
Just tested this and it does not work. The URL needs to be `choice.php?whichchoice=720&option=1&pwd&plant=${this.id}`.

horrible-little-slime:
I am going to become the joker

PR #289, opened 2022-05-07
Create CrimboShrub.ts

MrFizzyBubbs:


src/resources/2014/CrimboShrub.ts
```diff
@@ -0,0 +1,85 @@
...
+import { have as have_ } from "../../lib";
+import { get } from "../../property";
+import { $familiar, $item } from "../../template-string";
+
+export function have(): boolean {
+  return have_($familiar`Crimbo Shrub`);
+}
+
+type Topper = "Muscle" | "Mysticality" | "Moxie";
+type Lights = "Prismatic" | "Hot" | "Cold" | "Stench" | "Spooky" | "Sleaze";
```
horrible-little-slime, line 11: Switch this to `Prismatic Damage` and such in order to match the associated NumericModifier

src/resources/2014/CrimboShrub.ts
```diff
@@ -0,0 +1,85 @@
...
+type Garland = "HP Regen" | "PvP" | "Blocking";
```
horrible-little-slime, line 12:
```suggestion
type Garland = "HP Regen" | "PvP Fights" | "Blocking";
```
Don't actually commit that suggestion, because of what I'm about to say

src/resources/2014/CrimboShrub.ts
```diff
@@ -0,0 +1,85 @@
...
+type Gifts = "Yellow Ray" | "Red Ray" | "Gifts";
+
+function topperChoice(topper: Topper) {
```
horrible-little-slime, line 15: Instead of defining the types directly as union strings, and having these helper functions, instead do this:

```
const toppers = {
  "Muscle": 1;
  "Mysticality": 2;
  "Moxie": 3;
  } as const;
  type Topper = keyof typeof toppers;

**Review by horrible-little-slime — CHANGES_REQUESTED** [review]
I think I also gave this advice on discord, but I thought i'd be more precise here

**Comment by horrible-little-slime** [comment]
One final thing: export this from index.ts

File: src/resources/2014/CrimboShrub.ts [hunk 4]
```diff
@@ -42,13 +42,43 @@ const Gifts = {
...
+  Stench: "Stench Damage",
+  Spooky: "Spooky Damage",
+  Sleaze: "Sleaze Damage",
+  HP: "HP Regen",
+  PvP: "PvP Fights",
+  blocking: "Blocking",
+  yellow: "Yellow Ray",
+  meat: "Red Ray",
+  gifts: "Gifts",
+};
```
horrible-little-slime, line 61: do you not need `as const`?

src/resources/2014/CrimboShrub.ts
```diff
@@ -42,13 +42,43 @@ const Gifts = {
...
 ): boolean {
-  if (!have() || get("_shrubDecorated")) return false;
+  if (!have()) return false;
+  if (get("_shrubDecorated")) {
+    const decorations = [
+      get("shrubTopper"),
+      get("shrubLights"),
+      get("shrubGarland"),
+      get("shrubGifts"),
+    ].map((item) => Prefs[item as ShrubPref]);
```
horrible-little-slime, line 78: I feel like the use of `item` as your variable name here is potentially confusing

src/resources/2014/CrimboShrub.ts
```diff
@@ -42,13 +42,43 @@ const Gifts = {
...
+    return decorations === [topper, lights, garland, gifts];
```
horrible-little-slime, line 79: my understanding is that this will never be true, because anything more complex than a primitive is compared by reference rather than by value. You either want
```
[topper, lights, garland, gifts].every((x, i) => x === decorations[i])
```
Or you can use a utility function called, like, arrayEquals or something like that

horrible-little-slime (approved):

PR #290, opened 2022-05-09
Ponder orb

horrible-little-slime:
- Adds ponder() and have() functions to orb
 - Deletes currentPredictions function due to unreliability


src/resources/2021/CrystalBall.ts
```diff
@@ -1,38 +1,29 @@
...
-  return new Map(
-    predictions
-      .filter(
-        ([turncount]) =>
-          nonFreeCondition(turncount, myTurncount()) ||
-          (withFree && freeCondition(turncount, myTurncount()))
-      )
-      .map(([, location, monster]) => [location, monster])
-  );
+export function ponder(): Map<Location, Monster> {
```
docrostov, line 25: i think it is worth noting that once ponder was introduced i kind of thought this was the best way for libram to handle it so i am pretty overjoyed that we are finally just letting kol do it rather than trying to backend the comprehension of orb into libram. this is technically less efficient but i don't give the remotest shit, it works and is guaranteed to work regardless of any weird wanderer parsing or nonsense, thank you for doing this.

src/resources/2021/CrystalBall.ts
```diff
@@ -1,38 +1,29 @@
...
+  if (!have()) return new Map();
+  visitUrl("inventory.php?action=ponder", false);
+  return new Map(parsedProp());
```
docrostov, line 28: ... nevertheless, do we need a docstring here? here is my recommendation for one.

```
/**
 * @returns A list of all predictions currently active in an adventurer's miniature crystal ball, after visiting the "ponder" URL to refresh them.
 */
```

src/resources/2021/CrystalBall.ts
```diff
@@ -1,38 +1,29 @@
...
-      .filter(
-        ([turncount]) =>
-          nonFreeCondition(turncount, myTurncount()) ||
-          (withFree && freeCondition(turncount, myTurncount()))
-      )
-      .map(([, location, monster]) => [location, monster])
-  );
+export function ponder(): Map<Location, Monster> {
+  if (!have()) return new Map();
+  visitUrl("inventory.php?action=ponder", false);
```
docrostov, line 27: also i am still going to approve this because i think it's fine but i just want to register this here in the PR -- are we -positive- there is no real way libram will be calling this in the middle of some sort of chained NC/combat string, like the LOV tunnel? i expect not! you probably would need to put it in a postadventure script. but i am curious exactly -how- it would fuck up if you did that, and if there is room for an error catch here that simply returns the most recent predictions if a user messed up like that somehow.

docrostov (approved):
overall it looks fine. three minor things and one explicit bugcheck request, but i approve of it generally.

src/resources/2021/CrystalBall.ts
horrible-little-slime: _libram_ won't be calling this in the middle of some chained NC/combat string, but a player could. My general attitude is that if the product we produce is a drain cleaner, we aren't responsible for people who use it as an enema.

horrible-little-slime: Digging around, I have a way to check to see if you're in a multiFight, or if a choice follows fight, or whatever, but I don't see a way to determine if you're currently in combat

horrible-little-slime: :eyes:

src/lib.ts
```diff
@@ -688,6 +692,18 @@ export function getTodaysHolidayWanderers(): Monster[] {
     .flat();
 }
 
+/**
+ * Determines & returns whether or not we can safely call visitUrl(), based on whether we're in a fight, multi-fight, choice, etc
```
docrostov, line 696: good and useful function. docstring looks good. 

my only thought is: have you tested this? i see no obvious reason why this shouldn't work; to talk through the logic, a true in any of the conditions examined means that the X or Y or Z has spawned a true, which gets converted to false by the not operator. does current round default to 0? i could see it (possibly) defaulting to something weird instead but you probably have tested it.

docrostov (approved):
looks good. one small q re: testing but the logic looks right and non-broken, i just like seeing a test :p

src/lib.ts
horrible-little-slime: yes! it does

horrible-little-slime: And it works even if you win init

PR #291, opened 2022-05-10
Different flower paradigm

horrible-little-slime:


horrible-little-slime:
@MrFizzyBubbs how does this look as a new framework to start doing flower stuff on?

gausie:
reckon we can shift this paradigm?

PR #292, opened 2022-05-12
Fix mood MP reservation.

phulin:
In garbo, this was causing use of non-free restores every turn when building up turns of Pasta Eyeball in certain situations.

### DIFF
--- a/src/mood.ts
+++ b/src/mood.ts
@@ -11,7 +11,6 @@ import {
   haveSkill,
   hpCost,
   Item,
-  itemAmount,
   mallPrice,
   mpCost,
   myHp,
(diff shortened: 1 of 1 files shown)


src/mood.ts
```diff
@@ -70,28 +69,30 @@ export class OscusSoda extends MpSource {
...
+  }
+
   usesRemaining(): number {
-    return have($item`Kramco Sausage-o-Matic™`)
-      ? 23 - get("_sausagesEaten")
+    const maxSausages =
+      availableAmount($item`magical sausage`) +
+      availableAmount($item`magical sausage casing`);
+    return this.available()
+      ? Math.min(maxSausages, 23 - get("_sausagesEaten"))
```
horrible-little-slime, line 81: I'd like to see this as a clamp, in case _sausagesEaten gets misaligned somehow

horrible-little-slime (approved):
If this is tested, merge at your leisure

PR #293, opened 2022-05-19
robortender support

horrible-little-slime:


src/resources/2017/Robortender.ts
```diff
@@ -0,0 +1,55 @@
...
+  const phylum = target instanceof Monster ? target.phylum : target;
+  return phylumDrops.get(phylum) ?? $item`none`;
+}
+
+/**
+ * Determines the probability of getting a robortender drop based on number of drops received
+ * @param dropNumber The number of drops to assume you've already received; defaults to mafia's tracked amount
+ * @returns The probability of getting a robort drop
+ */
+export function dropChance(dropNumber = get("_roboDrops")): number {
```
docrostov, line 53: do we need to push the necessary mafia version for libram to ensure we've got the correct _roboDrops? i vaguely recall there being some small errors with the initial implementation. i suppose this isn't necessarily that big of an issue since this technically just floors at 20%, but might as well bring it up i guess.

src/resources/2017/Robortender.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * @returns Whether you have the Robortender in your terrarium/available
+ */
+export function have(): boolean {
+  return haveFamiliar(familiar);
+}
+
+const phylumDrops = new Map<Phylum, Item>([
+  [$phylum`Bug`, $item`pickled grasshopper`],
+  [$phylum`Constellation`, $item`bottle of anís`],
```
docrostov, line 19: i would recommend making anís into the item number reference rather than the item name. when log parsing there are consistently weird parsing errors when going cross-platform with special characters and i have a feeling that could (though is not guaranteed to) cause some heartburn here too.

docrostov (commented):
overall looks good to me. two minor things in here. do with them what you will.

src/resources/2017/Robortender.ts
horrible-little-slime: iirc libram doesn't have an associated mafia version, just the scripts that call it.

we probably should have one, though!

src/resources/2017/Robortender.ts
horrible-little-slime: I _think_ this has been fixed [as of recently](https://kolmafia.us/threads/java-bug-on-some-versions-of-java-weird-things-happen-to-strings-with-special-characters.27531/), but I'll investigate further

docrostov (approved):
thank you for changing the item id thing. approved.

src/resources/2017/Robortender.ts
lewismd13: I think that can be handled scriptside as needed. Though to a similar point, does new libram version automatically update property typing? otherwise that won't work, because get() will return a string for that prop, since it doesn't know what it is.

horrible-little-slime:
It will if I push the property typing thing!

PR #294, opened 2022-05-19
use reduce for CombatLoversLocket.findMonster in lieu of sort

horrible-little-slime:
This is a micro-optimization, but garbo might start calling this a lot more times than normal, so reducing the number of times we iterate over Monster.all() feels fortuitous

### DIFF
--- a/src/resources/2022/CombatLoversLocket.ts
+++ b/src/resources/2022/CombatLoversLocket.ts
@@ -74,10 +74,10 @@ export function reminisce(monster: Monster): boolean {
 }
 
 /**
- * Find a reminiscable monster that meets certain criteria and optionally maximizes a valuation function.
- * @param criteria A function for delineating which monsters are "fair game" for the search.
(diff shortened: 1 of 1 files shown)


docrostov:
the syntax changes seem fine so idk consider this a rubber stamp i guess

i will caveat that i do not think the description for this function does a particularly good job of walking through what it actually does. i would probably change it to something like this:
```/**
 * This function efficiently evaluates all of an adventurer's possibly reminiscable monsters, placing them through a filtering criteria and evaluating them based on a passed function.
 * @param criteria A filtering function for delineating which monsters are "fair game" for the search, like "is the monster free".
 * @param value A function for deciding which monsters are "better" than others.
 * @returns A singular monster that fulfills the criteria function and maximizes the value function.
 */```

docrostov (approved):
wow what a well written docstring, i wonder what handsome rogue editor was responsible for the minor polishing of the literary genius inherent in the O.G. docstring

PR #295, opened 2022-05-21
robortender feeding functions

horrible-little-slime:


src/resources/2017/Robortender.ts
```diff
@@ -55,3 +64,33 @@ export function dropFrom(target: Monster | Phylum): Item {
...
+  if (!pref) return [];
+  return pref
+    .split(",")
+    .filter((x) => x.trim())
+    .map((name) => toItem(name))
+    .filter((drink) => drinks.includes(drink));
+}
+
+/**
+ * @param beverage Feed the robortender a beverage of your choice.
```
docrostov, line 86: i'd change this docstring to:

```
/**
 * @param beverage A robortender-consumable drink of choice (i.e. Drive-By Shooting, Single Entendre, etc)
 * @returns A boolean; if true, the user's robortender has drunk that drink after execution. If false, it has not. This ALSO returns false if the user has not passed the function a robortender-consumable drink. If the user does not already have the beverage in their inventory, this function will not purchase the requested for you.
 */
```

docrostov (approved):
one docstring change but otherwise looks good to me.

PR #297, opened 2022-05-22
add toMonsterWrapper to account for kas's bug

horrible-little-slime:


horrible-little-slime:
fixing on mafia's end, like a god-fearing american.

PR #298, opened 2022-05-22
create default price function for ActionSource

horrible-little-slime:
Sometimes, mallPrice is an inappropriate way to determine how expensive a free run is. This is consistent with similar behavior we've implemented in diet.ts


src/actions/ActionSource.ts
```diff
@@ -102,6 +102,8 @@ function mergeConstraints(
  * A combat-based action resource in the game (e.g. a free run or free kill).
  */
 export class ActionSource {
+  static defaultPriceFunction = (item: Item) =>
+    mallPrice(item) > 0 ? mallPrice(item) : Infinity;
```
burningbman, line 106: This could be worth putting in a util file or something since it'll probably affect another area of code.

burningbman (approved):

PR #299, opened 2022-05-24
create permedSkills function; only perm skills that aren't already permed moreseo

horrible-little-slime:
The original intent of this was allow players to catch errors before hopping the gash, but it turns out to be very difficult to estimate the actual amount of karma you expect to have. Once we have a better solution that, I'll happily do it.


src/ascend.ts
```diff
@@ -24,6 +27,17 @@ export enum Lifestyle {
   hardcore = 3,
 }
 
+export function permedSkills(): Map<Skill, Lifestyle> {
+  return new Map(
+    Array.from(Object.entries(getPermedSkills)).map(
```
burningbman, line 32: getPermedSkills needs to be called

burningbman (changes requested):

src/ascend.ts
```diff
@@ -191,9 +206,16 @@ export function ascend(
...
-    for (const [skill, permLevel] of permSkills.entries()) {
-      if (permLevel !== Lifestyle.casual) {
-        const permText = permLevel === Lifestyle.hardcore ? "hcperm" : "scperm";
+    const currentPerms = permedSkills();
+    let karma = get("bankedKarma");
+    for (const [skill, prospectivePermLevel] of permSkills.entries()) {
+      const currentPermLevel = currentPerms.get(skill) ?? Lifestyle.casual;
+      if (prospectivePermLevel > currentPermLevel) {
+        const expectedKarma = 100 * (prospectivePermLevel - currentPermLevel);
+        if (karma < expectedKarma) continue;
```
burningbman, line 215: Do we want this? With some skills being super expensive, I think it might be best to just abort perming skills when there isn't enough karma to get them all.

horrible-little-slime: We don't want to abort from inside the gash, but we also can't accurately estimate karma until before we hop the gash. It's a topic worthy of debate, but it's not as clear-cut as you may assume.

s-k-z:
I don't think the error is accurate. It's not that one individual skill is too expensive, it's that the set of skills selected is too expensive.

burningbman (approved):

burningbman:
It makes sense to me to let the user decide if it's worth aborting during skill get. That'll let people more concerned with automation focus on never failing while others getting expensive skills can setup failsafes.

PR #308, opened 2022-06-11
restore more mp at once

horrible-little-slime:


horrible-little-slime:
This seems to be working as intended based on testing, the real question is "is working as I intended what others intend?"

PR #310, opened 2022-06-13
allow plural constants to escape commas

horrible-little-slime:


gausie (approved):
Do we want to add a `since`?

PR #313, opened 2022-06-30
Move from webpack to esbuild as a demonstration of the possibility

gausie:
Webpack: 3.00 seconds to make a 826KB build
Esbuild: 1.76 seconds to make a 705KB build

This also includes some bundle-size fixed I made because of bad imports from lodash applied to both cases


horrible-little-slime:
I don't think I can meaningfully review this, but I will say, I care more about build size than build time. It's nice to have quick builds when you're testing things, but it doesn't really matter that much at the end of the day. Sizes also don't matter, but at least that's something that is in a constant state of mattering slightly.

src/resources/2020/Guzzlr.ts
```diff
@@ -1,5 +1,5 @@
 import { Item, Location, mallPrice, runChoice, use, visitUrl } from "kolmafia";
-import { maxBy } from "lodash";
+import maxBy from "lodash/maxBy";
```
horrible-little-slime, line 2: Is this general spring cleaning, or is this something that we'll specifically want to start doing once we switch to esbuild?

gausie: We should be doing this now

gausie: Otherwise we massively inflate the build size

horrible-little-slime:
#414 414

PR #322, opened 2022-07-10
first pass implementation of telescope()

horrible-little-slime:


JamesDowney (approved):
I copied this into yorick and played around with it to test it, it worked as expected in every case I thought to test.

PR #323, opened 2022-07-11
Use file path in session save/load

MrFizzyBubbs:
@pstalcup


horrible-little-slime:
Got informal review from Mr. Killer

PR #324, opened 2022-07-13
we want things when they're cheap, not expensive

horrible-little-slime:


docrostov (approved):
i agree, phred. we want things when they are cheap, not expensive. good PR.

PR #326, opened 2022-07-14
upgrade kolmafia

horrible-little-slime:


horrible-little-slime:
Looks like kolmafia-js has committed the mortal sin of using `any` and has thus set the world aflame

PR #331, opened 2022-07-16
Initial Spacegate Support

Tokoeka:
Spacegate functions - getting info about current world, collecting your hazard equipment, dialing the gate, getting your vaccine


src/resources/2017/Spacegate.ts
```diff
@@ -0,0 +1,155 @@
...
+          return 1;
+        case $item`exo-servo leg braces`:
+          return 2;
+        case $item`rad cloak`:
+          return 3;
+        case $item`gate transceiver`:
+          return 4;
+        case $item`high-friction boots`:
+          return 5;
+      }
```
horrible-little-slime, line 107: This is repeated twice, which means it should precede the actual condition, and at the same time, given that you never invoke the function `num`, I'm pretty sure the visitUrls it's in aren't going to work.

Instead, consider defining a Map, or putting them all in an array and adding something to indexOf, or doing toInt to the equipment and subtracting something from it.

src/resources/2017/Spacegate.ts
```diff
@@ -0,0 +1,155 @@
...
+      case "Elemental Resistance":
+        return 1;
+      case "Broad-Spectrum":
+      case "Stats":
+        return 2;
+      case "Emotional":
+      case "Monster Level":
+        return 3;
+    }
+  };
```
horrible-little-slime, line 138: Ditto for this

horrible-little-slime (changes requested):

src/resources/2017/Spacegate.ts
```diff
@@ -0,0 +1,155 @@
+import { cliExecute, Item, visitUrl } from "kolmafia";
+import { get } from "../../property";
+import { $item, $items } from "../../template-string";
+
+export function have(): boolean {
+  return get("spacegateAlways") || get("_spacegateToday");
+}
+
+export function checkStatus(): void {
```
horrible-little-slime, line 9: `updateStatus` might be a more informative name for something with return type `void`

src/resources/2017/Spacegate.ts
```diff
@@ -0,0 +1,155 @@
...
+
+export function spants() {
+  return get("_spacegateSpant");
+}
+
+export function murderBots() {
+  return get("_spacegateMurderbot");
+}
+
+export function hazardEquipment(hazards: string): Item | Item[] {
```
horrible-little-slime, line 54: I think returning just Item[], even if that array is sometimes a singleton, is probably more useful?

src/resources/2017/Spacegate.ts
```diff
@@ -0,0 +1,155 @@
...
+  if (hazards.includes("irradiated")) equipment.push($item`rad cloak`);
+  if (hazards.includes("magnetic storms"))
+    equipment.push($item`gate transceiver`);
+  if (hazards.includes("high winds"))
+    equipment.push($item`high-friction boots`);
+  if (equipment.length === 2) {
+    return equipment[1];
+  } else {
+    return equipment;
+  }
```
horrible-little-slime, line 69:
```suggestion
const hazardEquipment = {
  "toxic atmosphere": $item`filter helmet`,
  "high gravity": $item`exo-servo leg braces`,
  "irradiated": $item`rad cloak`,
  "magnetic storms": $item`gate transceiver`,
  "high winds": $item`high-friction boots`,
};
return Object.entries(hazardEquipment).filter(([cue]) => hazard.includes(clue)).map(([, item]) => item)
```
or something like that?

As I write that, that specifically doesn't look great. but packing all of that information into a Map or an Object or something feels like a much more natural approach.

src/resources/2017/Spacegate.ts
Tokoeka: Makes sense, wasn't sure if it allowed singleton $item arrays

src/resources/2017/Spacegate.ts
```diff
@@ -0,0 +1,118 @@
...
+      case "Elemental Resistance":
+        return 1;
+      case "Broad-Spectrum":
+      case "Stats":
+        return 2;
+      case "Emotional":
+      case "Monster Level":
+        return 3;
+    }
+  };
```
horrible-little-slime, line 101: A few options here:

```
const num = 1 + Math.floor(["Rainbow", "Elemental Resistance", "Broad-Spectrum", "Stats", "Emotional", "Monster Level"].findIndex(choice) / 2);
```
(that is pretty bad)

```
let num: number;
    switch (choice) {
      case "Rainbow":
      case "Elemental Resistance":
        num = 1;
        break;
      case "Broad-Spectrum":
      case "Stats":
        num = 2;
        break;
      case "Emotional":
      case "Monster Level":
        num = 3;
        break;
    }
    ```
    
    or
    ```
    const nums = {
      "Rainbow": 1,
      "Elemental Resistance": 1,
      "Broad-Spectrum": 2,
      "Stats": 2,
      "Emotional": 3,
      "Monster Level": 3
      };
      const num = nums[choice];
      ```
      or finally
      ```
      const vaccines = ["Elemental Resistance", "Stats", "Monster Level"] as const;
      type Vaccine = typeof vaccines[number];
      ...
      const num = vaccines[choice] + 1;
      ```

Tokoeka: Thank You :)

horrible-little-slime (approved):

PR #335, opened 2022-07-23
Conspiracy Island Quest-state  & control-panel functions 

Tokoeka:
These are effectively TS versions of functions from Gausie/Dicts ash scripts. More than open to them being changed - this is just how I have managed to make them work - not sure if there is a neater way or if any of this would be considered extraneous for libram.

Also contemplating whether we need functions for checking if we have access to any of the shops: would appreciate opinions on this.


src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,286 @@
...
+    this.maximise = maximise;
+  }
+
+  currentQuest(): boolean {
+    if (get(this.questStateProperty) !== "unstarted") {
+      return true;
+    } else {
+      return false;
+    }
+  }
```
horrible-little-slime, line 75:
```suggestion
    return get(this.questStateProperty) !== "unstarted";

```

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,286 @@
...
+
+const radio = "place.php?whichplace=airport_spooky&action=airport2_radio";
+const omegapanel =
+  "place.php?whichplace=airport_spooky_bunker&action=si_controlpanel";
+
+export function have(): boolean {
+  return get("spookyAirportAlways");
+}
+
+class conspiracyQuestData {
```
horrible-little-slime, line 13: It is unclear to me why this is a Class

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,286 @@
...
+  ),
+] as const;
+
+export function available(): boolean {
+  return have() || get("_spookyAirportToday");
+}
+
+export function completedOneTimeQuests(): boolean {
+  for (const i in conspiracyQuests) {
+    const quest = conspiracyQuests[i];
```
horrible-little-slime, line 199:
```suggestion
  for (const quest of conspiracyQuests) {
```
While `for... in` iterates over the keys of a js Object, `for ... of` iterates over the values of an iterable

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,286 @@
...
+      return false;
+    }
+  }
+
+  return true;
+}
+
+export function haveQuest(): boolean {
+  for (const i in conspiracyQuests) {
+    const quest = conspiracyQuests[i];
```
horrible-little-slime, line 210:
```suggestion
  for (const quest of conspiracyQuests) {
```

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,286 @@
...
+    if (get(quest.questStateProperty) !== "") {
+      return true;
+    }
+  }
+  return false;
+}
+
+export function activeQuest(): string {
+  for (const i in conspiracyQuests) {
+    const quest = conspiracyQuests[i];
```
horrible-little-slime, line 220:
```suggestion
  for (const quest of conspiracyQuests) {
```

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,286 @@
...
+    if (get(quest.questStateProperty) !== "") {
+      return true;
+    }
+  }
+  return false;
+}
+
+export function activeQuest(): string {
```
horrible-little-slime, line 218: Is returning the name of a quest useful for people? Right now, `conspiracyQuests` isn't exported, so that name can't be used to access all of the valuable information stored in your quest instances.

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,286 @@
...
+      get(quest.questStateProperty) !== "unstarted" &&
+      get(quest.questStateProperty) !== "finished"
+    ) {
+      return quest.name;
+    }
+  }
+  return "";
+}
+
+export function getQuest(): void {
```
horrible-little-slime, line 231: `get` feels misleading here because we so often use it to mean "return the value of", instead of "obtain"--consider renaming this to something that more explicitly means that you're, like, obtaining and starting a quest

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,286 @@
...
+  if (get(`_questESp`) === "" && !haveQuest() && canInteract() && available()) {
+    visitUrl(radio);
+    runChoice(1);
+  }
+}
+
+export function questComplete(): boolean {
+  if (!haveQuest()) return false;
+  for (const i in conspiracyQuests) {
+    const quest = conspiracyQuests[i];
```
horrible-little-slime, line 241:
```suggestion
  for (const quest of conspiracyQuests) {
```

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,286 @@
...
+}
+
+export function turnInQuest(): void {
+  if (questComplete()) {
+    visitUrl(radio);
+    runChoice(1);
+  }
+}
+
+export function activateProtocol(protocol: number) {
```
horrible-little-slime, line 263: Make an enum or something for these protocols, so people can refer to them by useful names

src/resources/2014/ConspiracyIsland.ts
Tokoeka: are you querying the necessity of it at all, or just why it is a class instead of [other thing]? If the latter, because I dont know what else to make it (i tried to make it a "record"(?) but could not get that to work.

src/resources/2014/ConspiracyIsland.ts
Tokoeka: conspiracyQuests renamed to quests and exported

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,296 @@
...
+    for (const button of protocols) {
+      const name = button[0];
+      if (name === protocol) {
+        runChoice(button[1][2]);
+      }
+    }
+  }
+}
+
+export function activateOmega(): void {
```
horrible-little-slime, line 273: Can we get this to return a boolean of whether omega was successfully activated?

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,296 @@
...
+  ["VO-5", [$monster`creepy little girl`, $item`none`, 5]],
+  ["R-XPN", [$monster`creepy little girl`, $item`cuddly teddy bear`, 6]],
+  ["1862", [$monster`super-sized Cola Wars soldier`, $item`none`, 7]],
+  [
+    "1912",
+    [$monster`super-sized Cola Wars soldier`, $item`khaki duffel bag`, 8],
+  ],
+  ["0-0Z-E", [$monster`none`, $item`airborne mutagen`, 9]],
+  ["Ω", [$monster`none`, $item`none`, 10]],
+]);
```
horrible-little-slime, line 257: Make this an `as const` Object keyed by protocols; then do `type Protocol = keyof typeof protocols`, and make the input type of `activateProtocol` be `Protocol`

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,296 @@
...
+    get(quest.questStateProperty) ===
+      (quest.requiredItem === $item`none` ? "step1" : "step2")
+  );
+}
+
+export function hasActiveQuest(): boolean {
+  return hasQuest() && !questComplete();
+}
+
+export function turnInQuest(): void {
```
horrible-little-slime, line 236: Return a boolean of whether you successfully turned it in

src/resources/2014/ConspiracyIsland.ts
horrible-little-slime: maybe use typed objects instead of tuples. So it would look something like
```
{
  Juan: { monster: $monster`government scientist`, item: $item.none, choiceOption: 1 },
  3.14: { monster: $monster`government scientist`, item: $item`battery-powered drill`, choiceOption: 2 },
}
``` and so on

Typescript doesn't handle tuples very nicely, so using Objects instead tends to be nicer in most contexts. It also makes it a little easier to read; rather than doing `button[1][2]`, you would do `protocols[protocol].choiceOption`

File: src/resources/2014/ConspiracyIsland.ts [hunk 13]
```diff
@@ -0,0 +1,296 @@
...
+    this.questProgressProperty = questProgressProperty;
+    this.canUseWanderers = canUseWanderers;
+    this.canUseRunaways = canUseRunaways;
+    this.cleeshsafe = cleeshSafe;
+    this.questLocation = questLocation;
+    this.requiredItem = requiredItem;
+    this.maximise = maximise;
+  }
+
+  currentQuest(): boolean {
```
horrible-little-slime, line 72: I would rename this to `isCurrentQuest` if we're continuing to make this a class

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,296 @@
...
+
+const radio = "place.php?whichplace=airport_spooky&action=airport2_radio";
+const omegapanel =
+  "place.php?whichplace=airport_spooky_bunker&action=si_controlpanel";
+
+export function have(): boolean {
+  return get("spookyAirportAlways");
+}
+
+class QuestData {
```
horrible-little-slime, line 20: there's only one method on this class; it may be worth making this an object type instead.

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,296 @@
...
+
+export function hasQuest(): boolean {
+  return quests.some((q) => q.currentQuest());
+}
+
+export function activeQuest(): QuestData {
+  return quests.find((q) => q.currentQuest()) || BLANK_QUEST;
+}
+
+export function acceptQuest(): void {
```
horrible-little-slime, line 216: If there's a way for this to fail, I would make it return a boolean. I try to avoid `void`-returning functions in libram. It's certainly not a categorical rule, but if you can return useful information you should.

src/resources/2014/ConspiracyIsland.ts
```diff
@@ -0,0 +1,296 @@
...
+    .every((q) => get(q.questStateProperty) === "finished");
+}
+
+export function hasQuest(): boolean {
+  return quests.some((q) => q.currentQuest());
+}
+
+export function activeQuest(): QuestData {
+  return quests.find((q) => q.currentQuest()) || BLANK_QUEST;
+}
```
horrible-little-slime, line 214: Rather than having these be two separate functions, it may be nicer to have this be only the `activeQuest` function and have it return `Quest | undefined`. undefined is falsy, so you get the same information out of it.

horrible-little-slime (changes requested):
I'd like to see what @gausie has to say about this feedback; he may disagree with some of it, and you'll want to defer to him in any case where he does.

src/resources/2014/ConspiracyIsland.ts
gausie: Agreed

src/resources/2014/ConspiracyIsland.ts
gausie: The problem here is that we're asking people to write things like the omega unicode symbol. I think here is a time we need to make a enum-like object. Typescript is better at tuples these days anyway.

Tuples can be made with `const tuple = <T extends any[]>(...args: T): T => args` super easily

src/resources/2014/ConspiracyIsland.ts
gausie: Agreed

src/resources/2014/ConspiracyIsland.ts
gausie: Ambivalent on this

gausie:
@Tokoeka up to getting this merged still?

gausie:
It's been a year since the poke, closing.

PR #336, opened 2022-07-23
Dinseylandfill

Tokoeka:
First pass at dinseyland quest data, funfund acquisition & "can we fight Wart"


src/resources/2015/Dinseylandfill.ts
```diff
@@ -0,0 +1,316 @@
...
+
+export function have(): boolean {
+  return get("stenchAirportAlways");
+}
+
+export function available(): boolean {
+  return have() || get("_stenchAirportToday");
+}
+
+class questData {
```
gausie, line 22: Do we want to extract Quest into a class elsewhere and then make this extend it? Or prove it here and see for the future. Also please PascalCase class names

src/resources/2015/Dinseylandfill.ts
```diff
@@ -0,0 +1,316 @@
...
+}
+
+export function hasQuest(): boolean {
+  for (const quest of quests) {
+    if (quest.currentQuest()) {
+      return true;
+    }
+  }
+
+  return false;
```
gausie, line 192:
```suggestion
  return quests.some(q => q.currentQuest())
```

src/resources/2015/Dinseylandfill.ts
```diff
@@ -0,0 +1,316 @@
...
+    "",
+    "",
+    "",
+    "",
+    -1,
+    false,
+    $item`none`,
+    $location`none`
+  );
+}
```
gausie, line 214:
```suggestion
const BLANK_QUEST = new questData(
    "",
    -1,
    "",
    "",
    "",
    "",
    -1,
    false,
    $item`none`,
    $location`none`
  );

export function activeQuest(): questData {
  return quests.find(q => q.currentQuest()) || BLANK_QUEST;
}
```

src/resources/2015/Dinseylandfill.ts
```diff
@@ -0,0 +1,316 @@
...
+
+export function questComplete(): boolean {
+  const quest = activeQuest();
```
gausie, line 217: This is looping our quest list twice - just check if the active quest is the blank quest. Why are we using a blank quest in the first place? Why not just pass null and handle it.

src/resources/2015/Dinseylandfill.ts
```diff
@@ -0,0 +1,316 @@
...
+export function acceptQuest(priority: number | string): void {
+  const page: string = visitUrl(kioskUrl);
+  let choice = 6;
+  const at: number = indexOf(page, "Available Assignments");
+  if (at == -1) {
+    return;
+  }
+
+  const jobs = [
+    `Electrical Maintenance`,
```
gausie, line 234: Can't this list be extracted from `quests`?

src/resources/2015/Dinseylandfill.ts
```diff
@@ -0,0 +1,316 @@
...
+  if (typeof priority === "string") {
+    for (const quest of quests) {
+      if (quest.name === priority) {
+        prioritynum = quest.priority;
+        break;
+      }
+    }
+  } else {
+    prioritynum = priority;
+  }
```
gausie, line 253:
```suggestion
  const priorityNum = typeof priority === "string" ? (quests.find(q => q.name === priority)?.priority ?? 7) : priority;
```

src/resources/2015/Dinseylandfill.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  for (const job1 of jobs.slice(0, prioritynum)) {
```
gausie, line 255: This can also be rewritten more concisely. Also you want to find job1 first and then find job2, they don't need to be in the same loop.

gausie (changes requested):
Great work but needs some finessing!

src/resources/2015/Dinseylandfill.ts
Tokoeka: If someone else is willing to extract and extend it, I have no issue with that, but I'm happy to prove it here and wait and see if we want to do that when we come to implementing other realms

src/resources/2015/Dinseylandfill.ts
Tokoeka: I found this easier than trying to stop it complaining about null if I recall correctly.

src/resources/2015/Dinseylandfill.ts
Tokoeka: I think I've done a nicer job of it now?

gausie (approved):
Looking good! Thanks for getting all that done, and well done

PR #341, opened 2022-07-29
Get Crimbo Shrub decorations

MrFizzyBubbs:


src/resources/2014/CrimboShrub.ts
```diff
@@ -86,6 +86,8 @@ export function decorate(
   if (!have()) return false;
   if (get("_shrubDecorated"))
     return isDecoratedWith(topper, lights, garland, gifts);
+  if (!have_($item`box of old Crimbo decorations`))
+    useFamiliar($familiar`Crimbo Shrub`);
```
horrible-little-slime, line 90:
```suggestion
  if (!have_($item`box of old Crimbo decorations`)) {
    useFamiliar($familiar`Crimbo Shrub`);
  }
```

this is a pet peeve of bean's that is in garbo's linting rules, but apparently not libram's. Still probably better practice to do

PR #346, opened 2022-08-11
Export inventory operation

MrFizzyBubbs:
This would be useful for the profit tracking that will end up in grimoire.


MrFizzyBubbs:
Sorry for the messy commit history, it was an accident.

PR #347, opened 2022-08-11
Allow specifying a Witchess macro

MrFizzyBubbs:


src/resources/2016/Witchess.ts
```diff
@@ -51,5 +52,6 @@ export function fightPiece(piece: Monster): string {
   ) {
     throw new Error("Failed to start fight.");
   }
+  if (macro) return runCombat(macro.toString());
   return runCombat();
```
horrible-little-slime, line 56:
```suggestion
  return runCombat(macro?.toString());
```
Does this look better?

MrFizzyBubbs: Yes! I didn't know you could do that.

PR #349, opened 2022-08-18
Fix single step macros

s-k-z:
Single step macros are generated incorrectly. Fix by always appending a semicolon to a component and don't join with a semicolon.

Example error of what happens without a semicolon:
```
run_combat("skill 1234");

Macro override "skill 1234" returned void.
Round 1: Player executes a macro!
KoLmafia thinks it is round 2 but KoL thinks it is round 1
You're on your own, partner. (Macro Aborted: 37 instructions (In a row?!) executed without any actions being taken.).
```

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -166,7 +166,7 @@ export class Macro {
     );
     this.components = [
       ...this.components,
-      ...nextStepsStrings.filter((s) => s.length > 0).map((s)=>s+";"),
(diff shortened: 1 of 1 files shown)


gausie (approved):

phulin:
why is this necessary? seems better to keep semicolons off until we render to a string?

s-k-z:
That's another option.

s-k-z:
Added a regex to remove extra semicolons in case a user does
```
Macro.step(someOtherMacro.toString());
```

PR #351, opened 2022-08-19
Create Cartography.ts

MrFizzyBubbs:


src/resources/2020/Cartography.ts
```diff
@@ -0,0 +1,36 @@
...
+import { $skill } from "../../template-string";
+
+export const passive = $skill`Comprehensive Cartography`;
+export const skill = $skill`Map the Monsters`;
+
+export function have(): boolean {
+  return _have(passive);
+}
+
+export function mapMonster(location: Location, monster: Monster): void {
```
horrible-little-slime, line 22: We generally prefer returning a boolean to throwing; libram should only throw when it is catestrophic

src/resources/2020/Cartography.ts
```diff
@@ -0,0 +1,36 @@
...
+  useSkill($skill`Map the Monsters`);
```
horrible-little-slime, line 23: check `_monstersMapped` before running

src/resources/2020/Cartography.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (!get("mappingMonsters")) {
+    throw new Error("Failed to setup Map the Monsters.");
+  }
+  const turns = myTurncount();
+  while (get("mappingMonsters")) {
+    if (myTurncount() > turns) {
+      throw new Error("Map the Monsters unsuccessful?");
+    }
+    visitUrl(toUrl(location));
```
horrible-little-slime, line 32: This will have trouble with free zone intro adventures. I'm not sure the best way to handle that.

src/resources/2020/Cartography.ts
```diff
@@ -0,0 +1,36 @@
...
+    runChoice(1, `heyscriptswhatsupwinkwink=${monster.id}`);
```
horrible-little-slime, line 33: I would check `handlingChoice`, and add the additional parameter to runChoice to tell it that a combat is coming rather than separately runCombatting

src/resources/2020/Cartography.ts
horrible-little-slime: You can probably check if you're handlingChoice and check if your combat round is high and if not, go again

horrible-little-slime: I'd also check canAdventure(location)

horrible-little-slime (changes requested):

src/resources/2020/Cartography.ts
MrFizzyBubbs: The wiki says that runChoice always automates combats unless told not to

src/resources/2020/Cartography.ts
```diff
@@ -0,0 +1,48 @@
...
+    // Not in combat
+    if (myTurncount() > turns) {
+      throw new Error("Map the Monsters unsuccessful?");
+    }
+    visitUrl(toUrl(location));
+    if (handlingChoice() && lastChoice() === 1435) {
+      runChoice(1, `heyscriptswhatsupwinkwink=${monster.id}`);
+      return true;
+    } else {
+      runChoice(-1, false);
```
horrible-little-slime, line 44: mafia now will abort if you runChoice while not in a choice; if you adventure and you hit a combat directly, this will abort.

Which is fine actually I guess?

horrible-little-slime (approved):

PR #358, opened 2022-08-31
$path

gausie:
This does most of the move to $path, but we need to refactor `Paths.ts`. I think the easiest thing to do is make utils called canUseSkillBooks(path), hasALlPerms(path) etc etc


lewismd13 (approved):

src/ascend.ts
```diff
@@ -222,7 +222,7 @@ export function ascend(
   pet: Item | undefined = undefined,
   permSkills: Map<Skill, Lifestyle> | undefined = undefined
 ): void {
-  if (!path.classes.includes(playerClass)) {
+  if (playerClass.path !== path) {
```
horrible-little-slime, line 225:
```suggestion
  if (path === Path.none || (path.avatar && playerClass.path !== path))
```

horrible-little-slime: I am 80% sure this is what we want

gausie: oy vey

gausie: So maybe you want to do
```ts
if (path.avatar && playerClass.path !== path)
```
?

horrible-little-slime: That doesn't work. What if I try to ascend journeyman as ed?

I also think we should make sure you can't ascend as Path.none

PR #364, opened 2022-09-17
Nuclear autumn

Tokoeka:
Add functions for Fallout shelter actions in Nuclear Autumn


horrible-little-slime:
I think this could use some docstrings, but is an exciting start

Tokoeka:
> I think this could use some docstrings, but is an exciting start

what are docstrings when they are at home? this is a term I have never heard before i think

horrible-little-slime (approved):
In the future we can add return values to these if relevant (I don't know much about NA, but it's nice to return a thing if you can), but I see no reason not to approve it as-is

PR #365, opened 2022-09-18
First draft at modeable caching

horrible-little-slime:


gausie:
Looks good to me. Does it matter that some of those commands will equip the item and some wont?

horrible-little-slime:
We check haveEquipped, so unless they actively change the slot, it should be fine.

PR #368, opened 2022-09-22
Barrel Shrine

gausie:
- Add basic support for barrel shrine (smash party)


src/resources/2015/BarrelShrine.ts
```diff
@@ -0,0 +1,25 @@
...
+  return get("barrelShrineUnlocked");
+}
+
+export function smashParty() {
+  if (!have()) return;
+
+  const total = BARRELS.map((i) => availableAmount(i)).reduce(
+    (sum, q) => sum + q,
+    0
+  );
```
horrible-little-slime, line 17:
```suggestion
  const total = sum(BARRELS, availableAmount);
```

importing `sum` from utils. Extremely optional change.

src/resources/2015/BarrelShrine.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (total <= 0) return;
+
+  visitUrl("inv_use.php?pwd&whichitem=8568&choice=1");
+
+  for (let i = 0; i < total / 100; i++) {
+    runChoice(2);
```
horrible-little-slime, line 23: consider adding an `if (handlingChoice())` in front of this, just to make failure less loudly if it does ever fail.

horrible-little-slime (approved):
I had to think out loud about the `for` loop to make sure it worked when total was divisible by 100, but it feels like it does.

Two minor suggested tweaks but other than that, looks gold

PR #369, opened 2022-09-22
Upgrade mafia to 5.x

horrible-little-slime:


oxc:
You will definitely also have to update the peerDependency to allow version ^5.0.0

But can ^4.0.0 also be allowed still? What actually changed in kolmafia, why did they move to a new major version? If there are no breaking changes that affect libram (which seems to be the case if you have to change nothing else), the peer dependency should be `^4.0.0 || ^5.0.0`.

gausie:
> You will definitely also have to update the peerDependency to allow version ^5.0.0
> 
> But can ^4.0.0 also be allowed still? What actually changed in kolmafia, why did they move to a new major version? If there are no breaking changes that affect libram (which seems to be the case if you have to change nothing else), the peer dependency should be `^4.0.0 || ^5.0.0`.

It doesn't really matter in the end, you're not actually updating anything with breaking changes. We bumped funnily enough because our versioning scheme had a breaking change. It should stay on 5.x.x for a very long time now. I don't mind getting people to update their dependencies since there is zero migration time.

oxc:
In my project I also had to import all the proxy classes like Item, Effect, etc., whereas they were available globally before.
This can however be related to my bundler config, and/or already happened in a earlier version to which I did not update, so this might be unrelated to this PR.

gausie:
> already happened in a earlier version to which I did not update, so this might be unrelated to this PR.

This. In the future that should result in a major version change now

gausie (approved):

phulin:
Yeah, that change was made a while ago.

PR #372, opened 2022-09-27
Seasoning gives 1 adventure if difference between min and max is <= 1

Shiverwarp:
This is a total shot in the dark and boy does it look ugly


phulin:
I'd modify the MenuItem class instead to give access to these fields.

Shiverwarp:
> I'd modify the MenuItem class instead to give access to these fields.

I think this may be beyond me for now, this is really my first foray into coding beyond my little .ash scripts.

I see the comment of:

`  /**
   * Construct a new menu item, possibly with extra properties. Items in MenuItem.defaultOptions have intelligent defaults.
   * @param item Item to add to menu.
   * @param options.organ Designate item as belonging to a specific organ.
   * @param options.size Override item organ size. Necessary for any non-food/booze/spleen item.
   * @param options.maximum Maximum uses remaining today, or "auto" to check dailyusesleft Mafia property.
   * @param options.additionalValue Additional value (positive) or cost (negative) to consider with item, e.g. from buffs.
   * @param options.effect Effect associated with this menu item (pocket wish effect, sweet synthesis effect, pill keeper potion extension)
   * @param options.mayo Which mayo to use before item (ignored if mayo clinic is not installed or item is not a food)
   * @param options.note Any note to track information about item, to be used later
   */`

 in there but it all seems self referential and I don't seem to understand how these are actually populated

I don't mind trying to learn though, do you have a suggested starting point?

Like I would think something like options.advMin would equal menuItem.item.adventures.split(/[-–—]/)[0]
and options.advMax would be menuItem.item.adventures.split(/[-–—]/)[1]

phulin:
Ah, my mistake. Look at `expectedAdventures`.

src/diet/index.ts
```diff
@@ -61,6 +61,10 @@ function expectedAdventures(
     (itemType(item) === "booze" && item.notes?.includes("BEER"))
       ? 1.5
       : 1.3;
+  const seasoningAdventures =
+    item.adventures.split(/[-–—]/)[1] - item.adventures.split(/[-–—]/)[0] <= 1
```
s-k-z, line 65: What happens to items that do not have one of those three characters in the adventures field?
Why split based on those three characters, isn't a normal hyphen sufficient `-`?

Shiverwarp: I based the characters from earlier in the file here:

```
 if (item.adventures === "") return 0;
  const [min, recordedMax] = item.adventures
    .split(/[-–—]/)
    .map((s) => parseInt(s));
```

Bu actually I can just use those values and not split again, so made that change

s-k-z: I think you want `max` because `recordedMax` can be undefined.

Shiverwarp: Yeah haha, got a little ahead of myself, already changed it, thanks!

pstalcup (approved):

PR #376, opened 2022-10-04
Fix computation error in diet code.

phulin:


src/diet/index.ts
```diff
@@ -15,6 +15,7 @@ import {
   myPrimestat,
   mySpleenUse,
   npcPrice,
+  print,
```
pstalcup, line 18: remove unused import

pstalcup (changes requested):

pstalcup (approved):

PR #378, opened 2022-10-07
Autumnaton

horrible-little-slime:


src/resources/2022/AutumnAton.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export function zoneItems(): 3 | 4 | 5 {
+  return (3 + currentUpgrades().filter((u) => u.includes("arm")).length) as
+    | 3
+    | 4
+    | 5;
+}
+
+export function seasonalItems(): 1 | 2 {
+  return currentUpgrades().includes("base_blackhat") ? 2 : 1;
```
Tokoeka, line 112: i thought it was the cowcatcher  (collection prow) that increased seasonal items, with "base blackhat" increasing experience.

horrible-little-slime: Ahh, good catch, I wasn't sure which item corresponded to id.

src/resources/2022/AutumnAton.ts
```diff
@@ -0,0 +1,171 @@
...
+      ? target
+      : Array.isArray(target)
+      ? target.find((l) => locationsAvailable.includes(l))
+      : target(locationsAvailable);
+  if (!location) return null;
+  if (!locationsAvailable.includes(location)) return null;
+
+  runChoice(2, `heythereprogrammer=${location.id}`);
+  if (handlingChoice()) runChoice(3);
+  return currentlyIn();
```
s-k-z, line 75: Why not return location? What do you do if Mafia fails to get the location for whatever reason?

src/resources/2022/AutumnAton.ts
```diff
@@ -0,0 +1,171 @@
...
+  Location,
+  runChoice,
+  toLocation,
+  totalTurnsPlayed,
+  visitUrl,
+  xpath,
+} from "kolmafia";
+import { get } from "../../property";
+
+export const item = Item.get("Autumn-Aton");
```
s-k-z, line 15: Should be lowercase `autumn-aton`

src/resources/2022/AutumnAton.ts
horrible-little-slime: That's a good point; I was returning mafia's value because I want it to return falsy if it somehow fails. Which is more dangerous, a false negative or a false positive?

s-k-z: You can keep going with a false negative as long as KoL received the query to send the robot. KoLMafia just might not know where you sent it or when it's returning.

In the event of a false positive, you can check if you have the autumn-aton again. So that's quickly recoverable.

Perhaps there's a safer option if you grab the result from `runChoice` though I'm not sure what the contents of that will be here.

horrible-little-slime: Excellent point; edited

I think we can grab the result from runChoice but that's presumably what mafia is doing as well, so I think the failure case would probably mess with the both of us.

Tokoeka (approved):

PR #385, opened 2022-10-13
Add ifNot_ to macro

libraryaddict:


src/combat.ts
```diff
@@ -356,6 +398,20 @@ export class Macro {
...
 
+  /**
+   * Create a new macro with an "if !" statement.
+   * @param condition The BALLS condition for the if statement.
+   * @param ifFalse Continuation if the condition is false.
+   * @returns {Macro} This object itself.
+   */
+  static ifNot_<T extends Macro>(
+    this: Constructor<T>,
+    condition: Parameters<T["ifNot_"]>[0],
```
horrible-little-slime, line 409: Rather than using Parameters, can we name the type that we use in if and ifn't somewhere?

horrible-little-slime: Other than that, I'm into it.

libraryaddict: I'm not really that experienced in typescript to do that

horrible-little-slime:
Sniped you, I'm afraid

PR #390, opened 2022-10-17
Use git-publish-subdir-action to simplify workflow

s-k-z:


gausie:
Nice, let's try this out. I'm sure it's fine.

gausie (approved):

PR #391, opened 2022-10-18
Ignore trial items that are asking for what we cannot provide

libraryaddict:


phulin:
I don't think this logic is right, but I need to think about this more and reread the original code.

libraryaddict:
> This isn't right. A trial item that is later in the list could end up freeing the space needed.

What about removing the capacity check here?

https://github.com/loathers/libram/blob/main/src/diet/index.ts#L649

gausie:
@phulin any more thoughts? i know this is busy season

phulin:
I think the suggestion above is fine - the filtering should happen later.

gausie:
Ok. @libraryaddict can you sort that out?

PR #393, opened 2022-10-23
Create debug mode

horrible-little-slime:


src/logger.ts
```diff
@@ -1,37 +1,45 @@
 import { printHtml } from "kolmafia";
 
 const defaultHandlers = {
-  info: (message: string) => printHtml(`<b>[Libram]</b> ${message}`),
-  warning: (message: string) =>
+  info: (message: string): unknown => printHtml(`<b>[Libram]</b> ${message}`),
```
pstalcup, line 4: I think we should have a unified function these all call instead of repeating the printHtml snippet across all functions

src/logger.ts
```diff
@@ -1,37 +1,45 @@
...
-  setHandler(
-    level: "warning" | "info",
-    callback: (message: string) => unknown
-  ): void;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  setHandler(level: LogLevel, callback: (message: any) => unknown): void {
+  static get debugMode(): boolean {
+    return Logger.debug;
+  }
+  static set debugActive(debug: boolean) {
```
pstalcup, line 33: this should probably just set a level instead of having specific stuff for debug

pstalcup (changes requested):
Diet would be another great place to add debugging - i know it isn't directly related to what you did, so I don't think it needs to happen now.

src/logger.ts
horrible-little-slime: Should people be able to modify the root printing function directly, or keep modifying the individual printing functions?

src/logger.ts
horrible-little-slime: none < error < warning < message < debug, right? So if debug is on we want to show everything, if 'message' is on we want to show warnings and errors too, etc etc?

gausie:
@pstalcup has your review been addressed?

gausie (approved):
I declare @pstalcup overruled

PR #394, opened 2022-10-24
use retrieveItem in ronin

lewismd13:


horrible-little-slime:
This looks good to me, although at this point I want to basically make the fillTo function check `canInteract` and then call one of two non-exported functions, one for in-run and one for not.

That is probably needless, though.

Can you test this over the course of a loop-day?

horrible-little-slime (approved):

PR #396, opened 2022-11-05
Adjust ascension detection to work in compact character pane

ajcoppa:
The existing ascension detection code looks for "Astral Spirit" in the character pane, but when using KoL's built in "compact character pane" option, the class name is never shown:

Screen Shot 2022-11-05 at 12 05 39 PM

Instead, can we use this detection that checks whether Lvl. is followed by an image? I avoided coding in the actual image name in case KoL ever changes that for some reason.

Manually tested the ASH versions of these directly in the mafia CLI:
```
> ash contains_text(visit_url("charpane.php"), "Karma:")

Returned: true
```

Also added a more specific error string so it's clear what failure is happening if users encounter this in the future.

### DIFF
--- a/src/ascend.ts
+++ b/src/ascend.ts
@@ -218,11 +218,7 @@ function toMoonId(moon: MoonSign, playerClass: Class): number {
 }
 
 function isInValhalla(): boolean {
-  const charPaneText = visitUrl("charpane.php");
-  return (
-    containsText(charPaneText, "Lvl. <img src=") ||
-    containsText(charPaneText, "Level <img src=")
-  );
+  return containsText(visitUrl("charpane.php"), "Karma:");
 }
 
 /**
--- a/src/ascend.ts
+++ b/src/ascend.ts
@@ -271,11 +271,13 @@ export function ascend(
     throw new AscendError(illegalSkill);
   }
 
-  if (!containsText(visitUrl("charpane.php"), "Astral Spirit")) {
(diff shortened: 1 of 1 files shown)


ajcoppa:
Amusingly, while in Valhalla, you can't change your KoL options so I didn't have a way to manually test if this works correctly for the full character pane. It would be awesome if someone can confirm this text will work there too.

ajcoppa:
After some Discord feedback I tried using api.php to check the class in a more direct way, but unfortunately api.php doesn't work while in Valhalla and gets force-redirected to afterlife.php . Adjusted the string to just search for "Karma:" in the char pane instead – ready for review.

src/ascend.ts
```diff
@@ -271,11 +275,13 @@ export function ascend(
...
 
-  if (!containsText(visitUrl("charpane.php"), "Astral Spirit")) {
+  if (!isInValhalla()) {
     visitUrl("ascend.php?action=ascend&confirm=on&confirm2=on");
   }
-  if (!containsText(visitUrl("charpane.php"), "Astral Spirit")) {
-    throw new AscendError();
+  if (!isInValhalla()) {
+    throw new AscendError(
+      "Couldn't determine whether ascension completed successfully"
```
s-k-z, line 283: I wouldn't insert a new message here, the default is an important reminder.

s-k-z (changes requested):

s-k-z (approved):

PR #397, opened 2022-11-07
Add reagnimated gnome

MrFizzyBubbs:


src/resources/2012/ReagnimatedGnome.ts
```diff
@@ -0,0 +1,34 @@
...
+  return Object.values(bodyParts).find((part) => _have(part));
+}
+
+export function choosePart(part: BodyPart): boolean {
+  if (!have()) return false;
+  if (!chosenPart()) {
+    visitUrl("arena.php");
+    runChoice(4);
+  }
+  return chosenPart() === bodyParts[part];
```
horrible-little-slime, line 29: I didn't think they were meltables. You should be able to have multiples, which is problematic for this; instead I'd just do `return _have(bodyParts[part])`

src/resources/2012/ReagnimatedGnome.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export function expectedAdvsPerCombat(weight: number): number {
+  return 0.01 + weight / 1000;
```
horrible-little-slime, line 33:
```suggestion
  return 0.01 + 0.99 * weight / 1000;
```

You can only get one adventure per turn

horrible-little-slime (changes requested):

src/resources/2012/ReagnimatedGnome.ts
```diff
@@ -0,0 +1,33 @@
...
+export function choosePart(part: BodyPart): boolean {
+  if (!have()) return false;
+  if (_have(bodyParts[part])) return true;
+  visitUrl("arena.php");
+  runChoice(4);
+  return chosenParts().includes(bodyParts[part]);
+}
+
+export function expectedAdvsPerCombat(weight: number): number {
+  return Math.min(0.01 + weight / 1000, 1);
```
horrible-little-slime, line 32:
```suggestion
  return Math.min(0.01 + (weight / 1000) * 0.99, 1);
```

horrible-little-slime (changes requested):
Final thing

horrible-little-slime (approved):

PR #399, opened 2022-11-14
Fix ascension detection (again)

ajcoppa:
The previous iteration fixed it for compact character pane, but broke it for full character pane. This version looks for the infinity img tag used while in Valhalla.

I've manually tested this in the compact character pane but have not yet verified in full character pane – you can't change the option while in Valhalla. I can manually test tomorrow or someone else can confirm this works for full too.

### DIFF
--- a/src/ascend.ts
+++ b/src/ascend.ts
@@ -1,7 +1,6 @@
 import {
   Skill,
   Class,
-  containsText,
   eudoraItem,
   getCampground,
   getWorkshed,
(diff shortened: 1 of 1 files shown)


ajcoppa:
Just manually confirmed that this also works correctly for non-compact character pane. That actually uses the same inf_small.gif image as compact, so I could remove some of the wildcarding if desired.

s-k-z (approved):

PR #409, opened 2023-01-09
add fallbot uniques

horrible-little-slime:


src/resources/2022/AutumnAton.ts
```diff
@@ -169,3 +171,36 @@ export function zoneItems(): 3 | 4 | 5 {
...
+  underground: {
+    low: { index: 1, item: $item`autumn sweater-weather sweater` },
+    mid: { index: 5, item: $item`autumn dollar` },
+    high: { index: 8, item: $item`autumn years wisdom` },
+  },
+};
+
+function fallbotEnv(location: Location): "underground" | "indoor" | "outdoor" {
+  const env = location.environment as EnvironmentType;
+  return env === "underwater" ? "underground" : env;
```
Shiverwarp, line 196: After a small test, sending autumnaton with a cowcatcher to the Briny Deeps twice, first time it brought back ONLY an autumnFest ale, second time ONLY autumn years wisdom. 

These are the cowcatcher items, which seems to mean we don't get anything at all for sending it underwater.

It probably won't be relevant since we can't really get anything from underwater locations anyway, but just for accuracy

Shiverwarp (approved):
Looks good to me, shouldn't be too hard to adjust my stuff to use getUniques, just have to account for the null return possibility

PR #416, opened 2023-01-19
migrate garbo's maxBy to libram; update sum; move away from lodash's maxBy

horrible-little-slime:


phulin:
Do we just want to remove lodash entirely as a requirement? The only other use is in the source terminal code.

horrible-little-slime:
That's probably a good idea I'll look into it

gausie (approved):

PR #428, opened 2023-03-03
Create some additional utility functions

horrible-little-slime:
undelay and Delayed<T> exist as-is in grimoire right now, I just figured I'd wriggle them up the totem pole to libram


src/utils.ts
```diff
@@ -216,3 +216,39 @@ export function arrayEquals<T>(
...
+export type Delayed<T> = [T] extends [(...args: any) => any]
+  ? never
+  : T | (() => T);
+
+/**
+ * Used to collapse a Delayed<T> object into an entity of type "T" as represented by the object.
+ * @param delayedObject Object of type Delayed<T> that represents either a value of type T or a function returning a value of type T.
+ * @returns The return value of the function, if delayedObject is a function. Otherwise, this returns the original element.
+ */
+export function undelay<T>(delayedObject: Delayed<T>): T {
```
gausie, line 234: image

src/utils.ts
```diff
@@ -216,3 +216,39 @@ export function arrayEquals<T>(
...
+
+/**
+ * An object keyed by string type T, with values of S.
+ *  or contains a 'default' parameter to use as a fallback.
+ */
+export type Switch<T extends string, S> =
+  | Record<T, S>
+  | (Partial<{ [x in T]: S }> & { default: S });
+
+export function makeByXFunction<T extends string>(
```
gausie, line 246: Can you document this function? What is `makeByXFunction` meant to do?

The worse a function name is the better the docs gotta be.

gausie (changes requested):

src/lib.ts
```diff
@@ -807,3 +810,14 @@ export function telescope(): {
...
   };
 }
+
+/**
+ * @param options An object keyed by stat; it must either contain all stats, or have a `default` parameter.
+ * @returns The option corresponding to your primestat.
+ */
+export const byStat = makeByXFunction(() => myPrimestat().toString());
+/**
+ * @param options An object keyed by player class; it must either contain all classes, or have a `default` parameter.
```
gausie, line 820: Add a title to these docstrings and I think we're good

src/utils.ts
```diff
@@ -216,3 +216,44 @@ export function arrayEquals<T>(
   if (left.length !== right.length) return false;
   return left.every((element, index) => element === right[index]);
 }
+
+/**
+ * Type that extends any non-function entity--like a string, number, or array--into a itself and a no-input function that returns it.
+ * Used to interact with objects that could either be functions or static values.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Delayed<T> = [T] extends [(...args: any) => any]
```
gausie, line 225: I think unknown is better than any here? Maybe?

horrible-little-slime: you can't use the rest parameter for args with unknown, and once you're using any for that, why not just use the whole thing

horrible-little-slime: but I should document the dumb square brackets thing

gausie (approved):

PR #436, opened 2023-03-12
Prefer `sum` to `sumNumbers` with `map`

horrible-little-slime:
Imo `sum` with a callback function is much more legible (and I think technically a little more performant) than `sumNumbers` with a `map` function; `sumNumbers` should be reserved only for arrays that are already numerical


horrible-little-slime:
I'm not sure if that actually solves the problem insofar as people will
still shove mapped arrays into `sum` rather than `sumNumbers`. But it is an
elegant way of crunching functions together, defaulting the second
parameter to Number

PR #447, opened 2023-03-28
Add fakeUse function; add submitQuest to CCPayphone

horrible-little-slime:


gausie:
It looks like you could go further here and include a callback to then visit runChoice

PR #449, opened 2023-04-04
Genie bottle implementation

s-k-z:
Most importantly, this enables wishing of effects via description text. Useful for wishing for `[1701]Hip to the Jive` vs `[1872]Hip to the Jive` which are otherwise ambiguous.


gausie:
@s-k-z where are we with this? Close?

s-k-z:
I sat on it because I am unsure of what the behavior is on April Fools, and I haven't remembered to check

PR #452, opened 2023-04-12
Monkeypaw

horrible-little-slime:


src/resources/2023/CursedMonkeyPaw.ts
```diff
@@ -0,0 +1,118 @@
...
+                  !drop.quest && (type !== "c" || rate >= 1) // Remove random roll drops
+              )
+              .map(({ drop }) => drop)
+          )
+      )
+      .flat()
+  );
+}
+let _unwishableEffects: Effect[];
+function unwishableEffects(): Effect[] {
```
gausie, line 56: Can you add an explanatory comment here? This code isn't super readable to me at least

horrible-little-slime: done! took the opportunity to streamline it now that it's not a quick js one-liner

PR #460, opened 2023-04-16
Prune trivial exports

horrible-little-slime:


gausie:
I've always sort of liked this. Is there a performance or circular import or something reason to remove?

horrible-little-slime:
Just to put this here and not just in the ephemeral seas of discord: by removing these generically-named exports we reduce the number of dumb errors you can get by absentmindedly letting your IDE handle imports

PR #466, opened 2023-04-26
allow users to externally start timer for tasks and tests

horrible-little-slime:


MrFizzyBubbs:
I like it! It would be nice to have a `stopTimer()` for handling non-consecutive tasks or scripts erroring out.

horrible-little-slime:
That would unfortunately require nontrivially changing the way the timer works; we can put it in a future PR

PR #471, opened 2023-05-05
print "update kolmafia" message when a Type.get fails

horrible-little-slime:
This is more of a pitch than a PR, but it solves the problem of mafia constants in top-level code throwing errors before our `since` check is triggered

### DIFF
--- a/src/template-string.ts
+++ b/src/template-string.ts
@@ -1,4 +1,5 @@
 import {
+  abort,
   Bounty,
   Class,
   Coinmaster,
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
We can regex the stringified error if we're worried about this catching the wrong thing.

gausie:
I would change the message to be like "If you're sure this item exists you may need to update KoLmafa"

gausie (approved):
NOICE

PR #472, opened 2023-05-05
let people do $items.all()

horrible-little-slime:


gausie (approved):
curious that `as T[];` is necessary but i don't mind this function existing

horrible-little-slime:
> curious that `as T[];` is necessary but i don't mind this function existing

iirc, the various MafiaClasses inherit `all()` from `MafiaClass` itself, which just gives them a `MafiaClass[]` typing inherently.

I don't think there's a neat way to do, like, `all(): InstanceType<this>[]` but if there is, that would fix it

PR #479, opened 2023-05-12
Revert "commas are robots too"

horrible-little-slime:
Reverts loathers/libram#473

To be pushed once <https://github.com/kolmafia/kolmafia/pull/1697> is live


horrible-little-slime:
```> js familiarWeight(Familiar.get("comma chameleon"))

Returned: 100```
```
> We predicted the Breed More Collies test would take 2 turns, but it took 1 turns.
```
I'm going to go ahead and call this done, although it does sound like there might be more necessary mafia work

PR #484, opened 2023-05-24
improve maximizer handling of items with numbers & special characters in their names

horrible-little-slime:


Shiverwarp (commented):
Looks good to me according to what heex3 said on the mafia forums!

gausie (approved):

PR #488, opened 2023-05-29
export `Card` and `cards` from DeckOfEveryCard

horrible-little-slime:
Exporting the type is really the main goal here, but I think exporting the array has nonzero utility and because it's `as const`, doesn't present any dangers


gausie (approved):
That is fine, though `as const` is not a runtime feature, so it is problematic IN THEORY (it's fine). You could wrap it in `Object.freeze` to get the effect you want

horrible-little-slime:
Do we prefer Object.freeze or exporting a getter function

gausie:
> Do we prefer Object.freeze or exporting a getter function

I think neither is genuinely a fine option here, but I guess I prefer freeze

PR #493, opened 2023-06-06
Extensions to prop setting

gausie:
* Add increment and decrement functions for numeric property types
* Make set function return the set value (breaking change?)


horrible-little-slime (commented):
I think you can use generic typing more rigorously using the PropertyType type, but also I think there's a reason we stopped doing that so maybe I'm wrong

gausie:
Yes I think so too but can't remember why

src/property.ts
```diff
@@ -219,34 +219,80 @@ export function get(property: string, _default?: unknown): unknown {
...
+ * @returns New value
+ */
+export function increment(
+  property: NumericProperty,
+  delta = 1,
+  max = Infinity
+) {
+  const value = get(property);
+
+  if (!isNumericProperty(property)) return value;
```
horrible-little-slime, line 269: This feels unnecessary given that you already require `NumericProperty` typing

src/property.ts
```diff
@@ -219,34 +219,80 @@ export function get(property: string, _default?: unknown): unknown {
...
+ * @returns New value
+ */
+export function decrement(
+  property: NumericProperty,
+  delta = 1,
+  min = Infinity
+) {
+  const value = get(property);
+
+  if (!isNumericProperty(property)) return value;
```
horrible-little-slime, line 291: ditto

horrible-little-slime (approved):

gausie:
I missed your review @horrible-little-slime! Thanks for merging. I think having explicit checks is good since libram is used by non-TS scripts, especially via `jsl`.

PR #495, opened 2023-06-29
Horse support

horrible-little-slime:


src/resources/2017/Horsery.ts
```diff
@@ -0,0 +1,45 @@
...
+export function current(): Horse | null {
+  return (get("_horsery") || null) as Horse | null;
+}
+
+/**
+ * @param horse The horse to change to
+ * @returns Whether, at the end of all things, that is your horse
+ */
+export function changeHorse(horse: Horse): boolean {
+  cliExecute(`horsery ${horse}`);
```
Rinn, line 26: could check current() first and not run the command at all, and should probably check have() as well

horrible-little-slime: good shout

PR #496, opened 2023-06-30
Consider every familiar in the bjorn, even ones that don't have interesting drops

horrible-little-slime:


src/resources/2010/CrownOfThrones.ts
```diff
@@ -329,6 +329,13 @@ export const ridingFamiliars: readonly FamiliarRider[] = [
   },
 ] as const;
 
+const FULL_RIDING_LIST = [
+  ...ridingFamiliars,
+  ...Familiar.all()
+    .filter((f) => ridingFamiliars.some(({ familiar }) => familiar === f))
```
gausie, line 335: Is this meant to be an inverted condition?

horrible-little-slime: yes

burningbman (approved):

PR #506, opened 2023-08-08
Basic August support

horrible-little-slime:


src/resources/2023/AugustScepter.ts
```diff
@@ -0,0 +1,39 @@
...
+  $skills`Aug. 1st: Mountain Climbing Day!, Aug. 2nd: Find an Eleven-Leaf Clover Day, Aug. 3rd: Watermelon Day!, Aug. 4th: Water Balloon Day!, Aug. 5th: Oyster Day!, Aug. 6th: Fresh Breath Day!, Aug. 7th: Lighthouse Day!, Aug. 8th: Cat Day!, Aug. 9th: Hand Holding Day!, Aug. 10th: World Lion Day!, Aug. 11th: Presidential Joke Day!, Aug. 12th: Elephant Day!, Aug. 13th: Left/Off Hander's Day!, Aug. 14th: Financial Awareness  Day!, Aug. 15th: Relaxation Day!, Aug. 16th: Roller Coaster Day!, Aug. 17th: Thriftshop Day!, Aug. 18th: Serendipity Day!, Aug. 19th: Honey Bee Awareness Day!, Aug. 20th: Mosquito Day!, Aug. 21st: Spumoni Day!, Aug. 22nd: Tooth Fairy Day!, Aug. 23rd: Ride the Wind Day!, Aug. 24th: Waffle Day!, Aug. 25th: Banana Split Day!, Aug. 26th: Toilet Paper Day!, Aug. 27th: Just Because Day!, Aug. 28th: Race Your Mouse Day!, Aug. 29th: More Herbs\, Less Salt  Day!, Aug. 30th: Beach Day!, Aug. 31st: Cabernet Sauvignon  Day!`
+);
+
+const today = () => Number(todayToString()) % 100;
+
+/**
+ * @returns Today's august scepter skill
+ */
+export function todaysSkill(): Skill {
+  return SKILLS[today() - 1];
```
Tokoeka, line 23: Is there a reason for doing it this way rather than just referring to the skillnumber minus a constant (7451)? Would mean we dont have to list out all the skill names, which feels neater to me?

horrible-little-slime: Yeah we can do that. I still think we should export the constant.

Because you're doing arithmetic, you'll want to do like `toSkill((7451 + today()).toFixed(0))` because of dumb int vs float stuff

burningbman (approved):

PR #514, opened 2023-08-24
add clan.haveInRupus function

Tokoeka:
add fucntion to check if an individual piece of clan  rumpus room furniture is in the rumpus room

### DIFF
--- a/src/Clan.ts
+++ b/src/Clan.ts
@@ -4,6 +4,7 @@ import {
   cliExecute,
   getClanId,
   getClanName,
+  getClanRumpus,
   getPlayerId,
   Item,
   Monster,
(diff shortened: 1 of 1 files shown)


horrible-little-slime (commented):
I can see why this is in `Clan.ts`, but I'm not sure it actually belongs here if it isn't using Clan-hopping support. We have some analogous functions for the campground in `lib`, it may be better to put it there.

The big thing is you have essentially made this an instance method of the `Clan` class that doesn't ever use any attributes of the instance or features of the class itself.

gausie:
Agreed, move to lib. Mafia is already caching this so there's no need to add to the class.

gausie:
@Tokoeka I'll mark this as stale in one week unless indicated otherwise

gausie:
It's been 2 months! Closing

PR #522, opened 2023-09-13
refactor `getBanishedMonsters`

horrible-little-slime:
most of this is aesthetic, but I did switch the two `break`s in the for loop to `continue`s, which I think is safer

still doesn't account for turns? that is probably fine

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -462,28 +462,30 @@ export function getBanishedMonsters(): Map<Item | Skill, Monster> {
   const result = new Map<Item | Skill, Monster>();
 
   for (const [foe, banisher] of banishes) {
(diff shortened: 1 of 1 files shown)


gausie:
I'm afraid this piece of code is an extremely good candidate for a unit test.

gausie:
> > I'm afraid this piece of code is an extremely good candidate for a unit test.
> 
> Do we have an infrastructure that allows for that?

Yes! Thanks @oxc 😆

gausie:
@horrible-little-slime we now have quite a good testing infrastructure that should be easily copied

gausie:
Happy Birthday!!!

PR #530, opened 2023-10-02
Ascend object parameters and kolGender param

soolar:
Per some discussion earlier on discord, this PR switches the `ascend` function to take an object containing its parameters instead of a lengthy list of parameters you have to remember the order of and sub in undefined for to default and so on. While I was at it, I added the KolGender enum and support for `defaultGenderOverride` preference to specify what gender to ascend as, instead of always going female. It still defaults to female for those who don't specify something though.

THIS IS A BREAKING CHANGE. It will require a version update. I asked if I should keep the old version around to prevent breaking things that update, and was told it would be acceptable to do it this way, and so I did. Not sure if I'm supposed to bump the version as part of this PR or if that happens in a follow up change, so I haven't done it yet.

This is my first libram PR, so let me know if I missed any steps. I did run formatting and linting, at least.

### DIFF
--- a/src/ascend.ts
+++ b/src/ascend.ts
@@ -235,6 +235,7 @@ function isInValhalla(): boolean {
  * @param params.path path of choice, as a Path object--these exist as properties of Paths
  * @param params.playerClass Your class of choice for this ascension
  * @param params.lifestyle 1 for casual, 2 for softcore, 3 for hardcore. Alternately, use the Lifestyle enum
+ * @param params.kolGender Unfortunately limited to male or female (as strings). Defaults to female, or the value of the defaultGenderOverride preference, if it is set.
  * @param params.moon Your moon sign as a string, or the zone you're looking for as a string
  * @param params.consumable From the astral deli. Pick the container item, not the product. Defaults to astral six-pack, provide $item`none` for nothing.
  * @param params.pet From the astral pet store.
@@ -246,6 +247,7 @@ export function ascend({
   path,
   playerClass,
   lifestyle,
+  kolGender,
   moon,
   consumable,
   pet,
@@ -254,6 +256,7 @@ export function ascend({
   path: Path;
   playerClass: Class;
   lifestyle: Lifestyle;
+  kolGender?: "male" | "female";
   moon: InputMoonSign;
   consumable?: Item;
   pet?: Item;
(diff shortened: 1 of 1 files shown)


src/ascend.ts
```diff
@@ -229,36 +234,59 @@ function isInValhalla(): boolean {
...
+  playerClass: Class;
+  lifestyle: Lifestyle;
+  kolGender?: KolGender;
+  moon: InputMoonSign;
+  consumable?: Item;
+  pet?: Item;
+  permOptions?: { permSkills: Map<Skill, Lifestyle>; neverAbort: boolean };
+}): void {
+  const DEFAULT_VALUES = {
+    kolGender:
```
horrible-little-slime, line 262: kol stores gender as an int; `1` for male and `2` for female. Passing this as-is into the URL should cause an error

src/ascend.ts
```diff
@@ -229,36 +234,59 @@ function isInValhalla(): boolean {
...
-export function ascend(
-  path: Path,
-  playerClass: Class,
-  lifestyle: Lifestyle,
-  moon: InputMoonSign,
-  consumable: Item | undefined = $item`astral six-pack`,
-  pet: Item | undefined = undefined,
-  permOptions?: { permSkills: Map<Skill, Lifestyle>; neverAbort: boolean }
-): void {
+export function ascend(inputValues: {
```
horrible-little-slime, line 251: I know that you got this name from me, but we should probably call this `options` instead of `inputValues`

src/ascend.ts
```diff
@@ -229,36 +234,59 @@ function isInValhalla(): boolean {
...
+  path: Path;
+  playerClass: Class;
+  lifestyle: Lifestyle;
+  kolGender?: KolGender;
+  moon: InputMoonSign;
+  consumable?: Item;
+  pet?: Item;
+  permOptions?: { permSkills: Map<Skill, Lifestyle>; neverAbort: boolean };
+}): void {
+  const DEFAULT_VALUES = {
```
horrible-little-slime, line 261: ditto, but with `DEFAULT_OPTIONS`

src/ascend.ts
```diff
@@ -229,36 +234,59 @@ function isInValhalla(): boolean {
...
     throw new AscendError(playerClass);
   }
   if (path.id < 0) throw new AscendError(path);
 
   const moonId = inputToMoonId(moon, playerClass);
   if (moonId < 1 || moonId > 9) throw new Error(`Invalid moon ${moon}`);
+
   if (
-    consumable &&
+    consumable !== $item`none` &&
```
horrible-little-slime, line 289: just add `none` to the array below, I think

src/ascend.ts
```diff
@@ -267,7 +295,7 @@ export function ascend(
   }
 
   if (
-    pet &&
+    pet !== $item`none` &&
```
horrible-little-slime, line 298: ditto

src/ascend.ts
soolar: The enum stores them as such, so it should be fine shouldn't it?

horrible-little-slime: dang I could have sworn a minute ago these were plain strings. Yeah this is fine

soolar: You're not crazy, they were initially, but not by the time I actually put up the PR, I think? Were you looking at it when I pushed the branch before I put up the PR maybe? Although at that point I believe I did have it converting to 1 or 2 in the url, just further below.

horrible-little-slime (approved):

PR #533, opened 2023-10-12
updateOnLocationChanged for Outfit

Shiverwarp:
This is useful for item farming scripts where zones can have modifiers on them (Florist friars etc) and we don't want to overcap item drop.

Default is false because I think for the most part in KoL this isn't that relevant

### DIFF
--- a/src/maximize.ts
+++ b/src/maximize.ts
@@ -352,7 +352,7 @@ function checkCache(
 
   if (options.updateOnLocationChange && myLocation() !== entry.location) {
     logger.warning(
-      "Equipment found in maximize cache but familiar is different."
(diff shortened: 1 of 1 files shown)


horrible-little-slime:
Is this the best behavior for handling location changes?
Do we want it to add location to the cache key? or is this better behavior?

gausie:
I think it has to go [here](https://github.com/loathers/libram/blob/bb888f44dc492166624a3eb2fae9af3192352693/src/maximize.ts#L611-L617) for that

gausie:
@Shiverwarp I'll close this as stale in 1 week unless otherwise indicated

PR #534, opened 2023-10-15
ponder less frequently

horrible-little-slime:
I believe there are zero cases where this leads to bad stuff.

Right now, I use both `myTotalTurnsSpent` and `totalTurnsPlayed` in an attempt to obviate all edge cases. I'm not 100% sure if `myTotalTurnsSpent` counts turns spent crafting (and I'm not even sure if I know whether it _should_) and probably other stuff, so we also use `totalTurnsPlayed`

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "husky": "^4.3.6",
     "java-parser": "^1.4.0",
     "jest": "^27.1.0",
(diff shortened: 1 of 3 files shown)


libraryaddict (approved):
I think there's exceptionally few edge cases this shouldn't cover, could always revisit if someone does report. Which is unlikely.

PR #537, opened 2023-10-24
add `turnsSavedBy` method to the CS class

horrible-little-slime:


src/challengePaths/2015/CommunityService.ts
```diff
@@ -60,11 +61,22 @@ const visitCouncil = () => visitUrl("council.php");
...
+  modifier: NumericModifier,
+  ...effects: Effect[]
+): number {
+  const newEffects = effects.filter((e) => !have(e));
+  return (
+    numericModifier(modifier) +
+    sum(newEffects, (effect) => numericModifier(effect, modifier))
+  );
+}
+
```
s-k-z, line 74: This needs special-casing for any list of effects that includes Steely-eyed squint or Bow-legged swagger because:
```
numericModifier($effect`Steely-Eyed Squint`, "Item Drop")

Returned: 0.0

numericModifier($effect`Bow-legged swagger`, "Weapon Damage")

Returned: 0.0
```
This appears to always be so.

s-k-z: Additionally, numericModifier results don't appear to be multiplied if things like that are active.

s-k-z (approved):

PR #539, opened 2023-11-01
add GuidetoBurningLeaves

Ignose:
Offers basic leave support; GuidetoBurningLeaves.have(), GuidetoBurningLeaves.numberOfLeaves(), and GuidetoBurningLeaves.burnLeaves().

Does not include any of the specially spaded leaf burns as constants (to add), nor any function to take a fire nap.


src/resources/2023/GuidetoBurningLeaves.ts
```diff
@@ -0,0 +1,51 @@
...
+  itemAmount,
+  visitUrl,
+} from "kolmafia";
+import { haveInCampground } from "../../lib";
+import { $item, $monster } from "../../template-string";
+
+
+const item = $item`A Guide to Burning Leaves`;
+
+export const specialLeaves: Map<Number, Item> = new Map([
```
horrible-little-slime, line 13: You can use `Map<number, Item | Monster>` if you want to include the monsters in this map.

horrible-little-slime: That being said, I would maybe use an object given that our keys are numbers.

Also worth noting: we never want to use the capitalized version of primitive types; only `number`, `boolean`, `string`, never `Number`, `Boolean` or `String`

Ignose: So, I'll submit what happens when I try this (made an object originally), it makes vscode unhappy.

s-k-z:
I would consider renaming this to either BurningLeaves or PileOfBurningLeaves because *A Guide to Burning Leaves* is the packaged item name.

src/resources/2023/BurningLeaves.ts
```diff
@@ -0,0 +1,64 @@
+import { Item, Monster, itemAmount, visitUrl } from "kolmafia";
+import { haveInCampground } from "../../lib";
+import { $item, $monster } from "../../template-string";
+import { get } from "../../property";
+
+const item = $item`A Guide to Burning Leaves`;
+
+type Thing = Item | Monster;
```
horrible-little-slime, line 8: I think this raises more questions than it answers, and I'd just put use the union explicitly when it matters

src/resources/2023/BurningLeaves.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const specialLeaves: Map<Thing, number> = new Map([
```
horrible-little-slime, line 10:
```suggestion
export const specialLeaves: Map<Item | Monster, number> = new Map([
```

src/resources/2023/BurningLeaves.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/**
+ * @param leaves Burns the desired number of leaves
+ */
+export function burnLeaves(leaves: number): void {
+  if (leaves > numberOfLeaves()) {
+    visitUrl(`choice.php?pwd&whichchoice=1510&leaves=${leaves}`);
+  }
+}
```
horrible-little-slime, line 48: It's good to have functions return a value when they can. Here, we could use a `boolean` to represent success; so something like this: 
```suggestion
export function burnLeaves(leaves: number): boolean{
  if (leaves > numberOfLeaves()) {
    visitUrl(`choice.php?pwd&whichchoice=1510&leaves=${leaves}`);
    return true;
  }
  return false;
}
```

src/resources/2023/BurningLeaves.ts
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * @param th determines which thing (Monster or Item) we want to burn leaves for
+ * @returns The whether we can acquire the item or fight the monster in question
+ */
+export function canBurnFor(th: Thing): boolean {
```
horrible-little-slime, line 54:
```suggestion
export function canBurnFor(th: Monster | Item): boolean {
```

src/resources/2023/BurningLeaves.ts
```diff
@@ -0,0 +1,65 @@
...
+  return haveInCampground(item);
+}
+
+/**
+ * @returns The number of leaves we have remaining
+ */
+export function numberOfLeaves(): number {
+  return itemAmount($item`inflammable leaf`);
+}
+
```
s-k-z, line 38:
```suggestion

function visitLeaves() {
  visitUrl("campground.php?preaction=leaves");
}

```
Need to prepend something like this each time before querying choice.php. Otherwise you won't be in the choice adventure. Used like:
```ts
visitLeaves();
visitUrl(`choice.php?pwd&whichchoice=1510&leaves=${leaves}`);
```

package.json
```diff
@@ -53,7 +53,7 @@
     "husky": "^4.3.6",
     "java-parser": "^1.4.0",
     "jest": "^27.1.0",
-    "kolmafia": "^5.27647.0",
+    "kolmafia": "^5.27668.0",
```
gausie, line 56: Peer dependency also needs updated

s-k-z:
You don't need to check `instanceof `. What you get from burnItem will be `number | undefined` and then checking that it's not undefined will narrow it to `number`. So, both Items and Monsters behave the same way.

src/resources/2023/BurningLeaves.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+  return false;
+}
+
+/**
+ * @param th determines which thing (Monster or Item) we want to burn leaves for
+ * @returns Whether we can burn for that thing; or burns for that thing
+ */
+export function burnFor(th: Monster | Item): boolean {
+  if (th instanceof Item) {
```
s-k-z, line 66: You don't need to check `instanceof `. What you get from burnItem will be `number | undefined` and then checking that it's not undefined will narrow it to `number`. So, both Items and Monsters behave the same way.

horrible-little-slime (approved):

src/resources/2023/BurningLeaves.ts
horrible-little-slime: There _was_ actually a typing issue here, but it had to do with having the `Map<Monster | Item, number>` as a type description for the variable instead of applying the type arg to the Map constructor, making typescript's difficulty with tuples come into play.

That being said, you're right that instanceof isnt' the best solution. But it was a weird one!

PR #543, opened 2023-11-10
Resolve circular imports

oxc:
There are two circular imports involving CrystalBall and CampAway. The first one is an easy fix, the latter requires MoonSign to be moved to its own module, which seems sensible anyway, though.


horrible-little-slime (approved):

oxc:
Oops. Apologies, apparently my IDE did not add the new file 🤦🏻

PR #544, opened 2023-11-10
Fix check for Jumping in the Flames

oxc:
Found another twisted predicate (sorry I missed that before). Also added a check whether there are any leaves to jump in, and a return value whether the jump was successful.

### DIFF
--- a/src/resources/2023/BurningLeaves.ts
+++ b/src/resources/2023/BurningLeaves.ts
@@ -65,11 +65,17 @@ function visitLeaves() {
 }
 
 /**
- * Checks whether you can, then jumps into the fire
+ * Checks whether you can, then Jumps in the Flames
+ * @returns Whether or not you jumped in the flames
  */
-export function jumpInFire(): void {
-  if (!get("_leavesJumped", false)) {
-    visitLeaves();
-    runChoice(2);
+export function jumpInFire(): boolean {
+  if (get("_leavesJumped")) {
+    return false;
   }
+  if (get("_leavesBurned") === 0) {
+    return false;
+  }
+  visitLeaves();
+  runChoice(2);
+  return get("_leavesJumped");
 }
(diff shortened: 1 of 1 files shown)


src/resources/2023/BurningLeaves.ts
```diff
@@ -65,11 +65,17 @@ function visitLeaves() {
...
-export function jumpInFire(): void {
-  if (!get("_leavesJumped", false)) {
-    visitLeaves();
-    runChoice(2);
+export function jumpInFire(): boolean {
+  if (get("_leavesJumped")) {
+    return false;
+  }
+  visitLeaves();
+  if (get("_leavesBurned") === 0) {
```
oxc, line 76: Should I rely on _leavesBurned always being correct, and move this check before the visitLeaves() call?

gausie: I think we can rely on it being accurate. Though considering the cost of an int comparison vs a call to KoL servers, we could just do it before AND after

oxc: Well, KoLmafia has to check the page once to get the correct count, so it's only correct if only KoLmafia has been used to manipulate it. But I guess that holds true for most of the properties.
If I check it before the visit, there will never be an update, so if it was not correct, we'll never know :)

oxc: I've moved the check to before the visitLeaves call. If the user has already burned some leaves, they will simply have to make KoLmafia aware of it by visiting the page

I have not added second check, because if KoLmafia already thinks that some leaves were burned, it is VERY unlikely that this is not the case. I can't think of a way to induce such a state.

horrible-little-slime (approved):

oxc:
Can we get this released maybe? The currently released implementation of BurningLeaves is broken in all methods beyond visiting the campground.

horrible-little-slime:
> Can we get this released maybe? The currently released implementation of BurningLeaves is broken in all methods beyond visiting the campground.

I was planning on cutting a new version after https://github.com/kolmafia/kolmafia/pull/2027 but will probably instead do so after https://github.com/loathers/libram/pull/549 , which is being looked at right now (it kicks ass)

PR #549, opened 2023-11-17
Properly encode message parameter in KMail.send

### LINKED ISSUE #505: `Kmail.send` breaks if messages include `&`
The current implementation of `send` will inject the `message` directly (without encoding) when constructing the url to invoke the send.

This means if the message contains an ampersand that is interpreted as being part of the URL (and the rest of the message is treated as part of the URL).  This results in unexpected behavior and is a potential security vulnerability (e.g. if someone incorporates user input / quotes in the construction of the message).

The most correct solution is to `urlEncode` the message before constructing the query string -- however, `visitUrl()` also encodes each piece of the query string by default (which, left unaddressed, would result in a double encoding).   To prevent this we would need to call `visitUrl` and pass `false` to bypass its encoding.  We can't do *that* safely unless we manually encode all the other query string parameters as well.

The way `send` is currently written has various query string parameters that are built across a few different locations, through a [generic send function](https://github.com/loathers/libram/blob/87e39f18e455de51dec98cb14faee88e25eee8f7/src/Kmail.ts#L68-L118) that partially constructs parameters via a provided callback.  Being able to safely ensure that all parts of the URL are encoded properly might warrant a small internal refactor of that logic.

oxc:
This is an attempt to Fix #505 

I've added a fetchUrl method that takes a path and a query parameter, and makes sure the parameters are properly encoded.

Using this method outside of KMail is not in the scope of this PR, but would probably be useful also in other places.

I'm happy to change the naming, if somebody can come up with a better suggestion.

To make sure I don't break anything, I've added a test suite for KMail.send and KMail.gift. The fetchUrl-change causes slight changes in the generated URLs, but this is expected and in fact desired. The changes applied to the test file in the second commit sound sensible to me, but please someone double check.

### DIFF
--- a/__tests__/Kmail.ts
+++ b/__tests__/Kmail.ts
@@ -1,0 +1,162 @@
+import { isGiftable, Item, visitUrl } from "kolmafia";
+import { $item, Kmail } from "../src";
+
+import mocked = jest.mocked;
+
+jest.mock("kolmafia");
+
+const USER = 11;
+
+const mockItem1 = $item`Jumbo Dr. Lucifer`;
+const mockItem2 = $item`worthless trinket`;
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+describe(Kmail.send, () => {
+  beforeEach(() => {
+    mocked(visitUrl).mockReturnValue(`<div>Message sent.</div>`);
+    mocked(isGiftable).mockReturnValue(true);
+  });
+
+  it("should send a kmail with a message", () => {
+    Kmail.send(USER, "Hello, world!");
+
+    expect(visitUrl).toHaveBeenCalledWith(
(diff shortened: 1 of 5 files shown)


oxc:
Small update, the default should be POST of course, I misread that in the visitUrl documentation

gausie:
I am so sorry this has not been reviewed. It rocks and I will look at it asap. Are you on the ASS discord btw?

gausie (approved):
This looks great. Thank you so much for the work. The testing framework is genuinely huge.

oxc:
> I am so sorry this has not been reviewed. It rocks and I will look at it asap. Are you on the ASS discord btw?

No, I'm not, but shoot me some details and I might try to join

PR #552, opened 2023-11-20
delete properties that didn't exist prior to the PM call

horrible-little-slime:


src/property.ts
```diff
@@ -385,6 +387,7 @@ export function withChoice<T>(
 }
 
 export class PropertiesManager {
+  private static EMPTY_PREFERENCE = Symbol();
```
gausie, line 390: Minor but I would add a descriptor string to this Symbol as good practice.

gausie (approved):
This is good to me. Have you tested it in Rhino. Both Rhino and symbol are weird, so who knows about their confluence.

PR #556, opened 2023-12-03
delete path.ts

horrible-little-slime:
I am pretty sure(?) this at the very least fixes the issue with `Path.get` when `jsl`ing your way into `ascend`


Rinn:
i love deleting code

Rinn (approved):

PR #581, opened 2024-03-25
support new combat flag "boringdarts"

horrible-little-slime:


gausie:
There is precedent for explaining each setting in the following comment

PR #592, opened 2024-04-23
Additional mimic check

Ignose:
Use Case: Say you're trying to write logic for Garbo to check how many KGE eggs you have, and whether you even have KGE eggs - without using them yet - so you can plan diet, plan your route, plan when to use Mimic to acquire eggs to begin with!

Well in that scenario, you'd need to know how many of your eggs can differentiate into KGEs. This solve that.


src/resources/2024/ChestMimic.ts
```diff
@@ -130,3 +132,18 @@ export function differentiate(
...
+ * @param monster The monster to differentiate your egg into; may behave weirdly with name collisions
+ * @returns How many of a Monster we can differentiate
+ */
+export function differentiableQuantity(monster: Monster): number {
+  if (!have_($item`mimic egg`)) return 0;
+  const regex = new RegExp(`${monster.name}\\s*(?:\\((\\d+)\\))?`, "i");
+  const page = examine($item`mimic egg`);
+  const match = page.match(regex);
+  if (!match) return 0;
+  return clamp(parseInt(match[1]), 0, availableAmount($item`mimic egg`)) || 1;
```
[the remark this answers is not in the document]
Ignose: Because if no number shows up, but the monster name shows up, you have exactly 1 egg with 1 monster.

src/resources/2024/ChestMimic.ts
```diff
@@ -130,3 +132,18 @@ export function differentiate(
...
+
+/**
+ * Check how many of a monster is available to differentiate into
+ *
+ * @param monster The monster to differentiate your egg into; may behave weirdly with name collisions
+ * @returns How many of a Monster we can differentiate
+ */
+export function differentiableQuantity(monster: Monster): number {
+  if (!have_($item`mimic egg`)) return 0;
+  const regex = new RegExp(`${monster.name}\\s*(?:\\((\\d+)\\))?`, "i");
```
horrible-little-slime, line 144: I would not make this regex case-insensitive, it makes this more prone to errors

horrible-little-slime: also, I have an extremely funny test case: do you know if it uses commas when it hits 4 digits?

Ignose: I have no idea! It would take 100 days to find out

horrible-little-slime (approved):

PR #598, opened 2024-04-27
Improve/fix Kmail parsing

oxc:
* correctly parse inside and outside note of gift message
* extract items and meat only from the actual attachments (not the text)
* don't swallow actually messages of valentine cards
* decode html entities in message
* use unixtime as source for the message date


gausie (approved):
Awesome

PR #608, opened 2024-05-03
update `differentiableQuantity` regex; update failure behavior; remove now-unnecessary warning

horrible-little-slime:


src/resources/2024/ChestMimic.ts
```diff
@@ -136,17 +136,14 @@ export function differentiate(
...
- * @param monster The monster to differentiate your egg into; may behave weirdly with name collisions
+ * @param monster The monster to differentiate your egg into
  * @returns How many of a Monster we can differentiate
  */
 export function differentiableQuantity(monster: Monster): number {
   if (!have_($item`mimic egg`)) return 0;
-  const regex = new RegExp(
-    `<!-- monsterid: ${monster.id}-->(?:\\s\\((\\d+)\\))?`
-  );
+  const regex = new RegExp(`<!-- monsterid:${monster.id} -->\\s\\((\\d+)\\)`);
```
s-k-z, line 144:
```suggestion
  const regex = new RegExp(`<!-- monsterid:${monster.id} --> \\((\\d+)\\)`);
```

No need in mixing whitespace literals and whitespace metacharacters before/after `-->` unless you expect it to dynamically change.

src/resources/2024/ChestMimic.ts
```diff
@@ -136,17 +136,14 @@ export function differentiate(
...
   const page = examine($item`mimic egg`);
   const match = page.match(regex);
-  if (!match) return 0;
-  if (!match[1]) return 1;
+  if (!match?.[1]) return 0;
   return clamp(Number(match[1]), 1, availableAmount($item`mimic egg`));
```
s-k-z, line 148:
```suggestion
  return Math.min(Number(match[1]) || 0, itemAmount($item`mimic egg`));
```
Should be itemAmount for that one guy (HolderOfSecrets)

s-k-z (changes requested):

src/resources/2024/ChestMimic.ts
s-k-z: Oops nevermind this won't work how I thought. Can't do `undefined[1]`

s-k-z (approved):

PR #615, opened 2024-05-11
chore: make pickX function accessible

oxc:
This is a counter-proposal to #614. Not sure what it's needed for exactly :)


gausie:
We could also do this but #614 is designed to improve the API of byClass, byStat etc

oxc:
As discussed on discord, there is a good use case for having an alternative source, and using `source` just as the default if not overridden.

PR #616, opened 2024-05-19
BALLS improvements

Rinn:
* while_ now accepts PreBALLSPredicate instead of just string for the condition
* repeat now optionally accepts PreBALLSPredicate
* the final repeat in trySkillRepeat now checks hasskill so it will break if the skill is no longer available
* make makeBALLSPredicate public and remove unneeded let variable


horrible-little-slime (approved):

horrible-little-slime:
improvema BALLS

PR #623, opened 2024-06-06
Add mini Kiwi Aioli to the diet

Ignose:


src/diet/index.ts
```diff
@@ -439,6 +447,15 @@ class DietPlanner<T> {
       helpers.push(this.whetStone);
     }
 
+    if (
+      this.aioli &&
+      itemType(menuItem.item) === "food" &&
+      this.mpa * menuItem.item.fullness >
+        Math.min(mallPrice($item`mini kiwi aioli`))
```
pstalcup, line 454: this should not use hte mall price, it should use the menuItem aioli.price

pstalcup (changes requested):

src/diet/index.ts
Ignose: Should Seasoning, Whetstone, etc. be updated as well?

pstalcup (approved):

PR #625, opened 2024-06-07
Fix modifier parsing

gausie:
- **Bump dependencies including java-parser, patch the incorrectly specified TS types**
- **Extract new modifier types**
- **Correct lint-staged command**

This is a breaking change as we no longer export a certain set of modifier types.


horrible-little-slime (approved):
Sad but acceptable

PR #633, opened 2024-08-01
Add you Robot support

Ignose:


src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,113 @@
...
+} from "kolmafia";
+import { have } from "../../lib.js";
+import { get } from "../../property.js";
+import { $path, $skill, $slot, $stat } from "../../template-string.js";
+
+export const path = $path`You, Robot`;
+
+type Upgrade = "robot_shirt" | "robot_energy" | "robot_potions";
+
+export class YouRobot {
```
gausie, line 18: Don't make classes that only have static functions, just export each of the functions. This module can act as the class. I'll add this as an eslint rule if I can find one.

All of these functions should have jsdocs

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,113 @@
...
+  static scrap(): number {
```
gausie, line 19: This seems unnecessary

gausie (commented):
Also get more creative with the do prefix pls

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,113 @@
...
+    return myRobotScraps();
+  }
+
+  static doScavenge(): void {
```
horrible-little-slime, line 23: Functions should return a meaningful value if they can, and this one absolutely can.

src/challengePaths/2021/YouRobot.ts
Ignose: Resolved both of these; good shout.

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+ * Spends energy at the chronolith to get 10 adventures
+ */
+export function visitChronolith(): void {
+  visitUrl("place.php?whichplace=scrapheap&action=sh_chronobo");
+}
+
+/**
+ * @returns how much energy we expect the next chronolith use to take
+ */
+export function expectedChronolithCost(): number {
```
gausie, line 42: It’s not clear from name or doc how this differs from expectedEnergyNextCollect

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+  return get("_chronolithNextCost");
+}
+
+/**
+ *
+ * @param stat the stat we want to boost using the statbot
+ */
+export function visitStatbot(stat: Stat) {
```
gausie, line 50: Visit says idempotent to me

src/challengePaths/2021/YouRobot.ts
Ignose: This covers how much the chronolith will cost, while expectedEnergyNextCollect covers how much energy you'll get next time you collect energy.

What would you prefer they be called?

src/challengePaths/2021/YouRobot.ts
Ignose: Are you in a You, Robot run?

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+export function scavenge(): void {
+  visitUrl("place.php?whichplace=scrapheap&action=sh_scrounge");
+}
+
+/**
+ * Spends one turn collecting energy in the scrapheap
+ */
+export function collectEnergy(): void {
+  visitUrl("place.php?whichplace=scrapheap&action=sh_getpower");
+}
```
horrible-little-slime, line 22: These two functions are very similar. They should have more similar docstrings.

I'd also have them return a `boolean` that essentially says "did we successfully do this"--maybe add checks for if the user is in YR and has an adventure. Alternately, you could return the total amount of energy or scavenge parts? There is a meaningful value you can return and you should return it.

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * @returns the energy cost for the next use of the chronolith
+ */
+export function expectedEnergyNextCollect(): number {
+  const raw = (25 + get("youRobotPoints")) * 0.85 ** get("_energyCollected");
+  return round(raw);
```
horrible-little-slime, line 29:
```suggestion
  return Math.round(raw);
```

No need to import the ash `round` function

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * Spends energy at the chronolith to get 10 adventures
+ */
+export function visitChronolith(): void {
+  visitUrl("place.php?whichplace=scrapheap&action=sh_chronobo");
+}
```
horrible-little-slime, line 37: This similarly should find a value to return, and should return it

src/challengePaths/2021/YouRobot.ts
horrible-little-slime: https://en.wikipedia.org/wiki/Idempotence

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ *
+ * @param stat the stat we want to boost using the statbot
+ */
+export function visitStatbot(stat: Stat) {
+  visitUrl("place.php?whichplace=scrapheap&action=sh_upgrade");
+  if (stat === $stat`Muscle`) runChoice(1);
+  else if (stat === $stat`Mysticality`) runChoice(2);
+  else if (stat === $stat`Moxie`) runChoice(3);
```
horrible-little-slime, line 54:
```suggestion
 const option = $stats`Muscle, Mysticality, Moxie`.indexOf(stat) + 1;
 if (!option) return;
 visitUrl("place.php?whichplace=scrapheap&action=sh_upgrade");
 runChoice(option);
```

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+    case "top":
+      return get("youRobotTop");
+    case "left":
+      return get("youRobotLeft");
+    case "right":
+      return get("youRobotRight");
+    case "bottom":
+      return get("youRobotBottom");
+  }
+}
```
horrible-little-slime, line 97:
```suggestion
function getPartId(slot: "Top" | "Left" | "Right" | "Bottom") {
  return get(`youRobot${part}`);
}
```

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ *
+ * @param which defines which section you want to swap from
+ * @param id defines the part id you want to swap to
+ * @returns whether the swap was successful
+ */
+export function doSwitchPart(
+  which: "top" | "left" | "right" | "bottom",
```
horrible-little-slime, line 106:
```suggestion
  slot: "Top" | "Left" | "Right" | "Bottom",
```

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+    `choice.php?whichchoice=1445&part=${which}&show=${which}&option=1&p=${id}`,
+  );
+  return getPartId(which) === id;
+}
+
+/**
+ * @returns whether current parts allow use of a familiar
+ */
+export function canUseFamiliar(): boolean {
+  return getPartId("top") === 2;
```
horrible-little-slime, line 122:
```suggestion
  return getPartId("Top") === 2;
```

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ *
+ * @param slot defines the slot you want to equip something to
+ * @returns whether that slot is available for equipment
+ */
+export function canUse(slot: Slot): boolean {
+  if (slot === $slot`familiar`) return getPartId("top") === 2;
```
horrible-little-slime, line 131:
```suggestion
  if (slot === $slot`familiar`) return canUseFamiliar();
```

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (slot === $slot`hat`) return getPartId("top") === 4;
```
horrible-little-slime, line 132:
```suggestion
  if (slot === $slot`hat`) return getPartId("Top") === 4;
```

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (slot === $slot`weapon`) return getPartId("left") === 4;
```
horrible-little-slime, line 133:
```suggestion
  if (slot === $slot`weapon`) return getPartId("Left") === 4;
```

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (slot === $slot`off-hand`) return getPartId("right") === 4;
```
horrible-little-slime, line 134:
```suggestion
  if (slot === $slot`off-hand`) return getPartId("Right") === 4;
```

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (slot === $slot`pants`) return getPartId("bottom") === 4;
```
horrible-little-slime, line 135:
```suggestion
  if (slot === $slot`pants`) return getPartId("Bottom") === 4;
```

src/challengePaths/2021/YouRobot.ts
horrible-little-slime: That being said, this name to me suggests it's a function where you input a _part_, and get the id number associated with that.

I'd consider providing Enums or something like that to allow people to refer to specific parts by name, instead of having to know their ids off the top of their head.

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ *
+ * @param which defines which section you want to swap from
+ * @param id defines the part id you want to swap to
+ * @returns whether the swap was successful
+ */
+export function doSwitchPart(
+  which: "top" | "left" | "right" | "bottom",
+  id: number,
```
horrible-little-slime, line 107: Variable names here aren't very informative at all--jsdoc helps, but isn't sufficient.

src/challengePaths/2021/YouRobot.ts
```diff
@@ -0,0 +1,139 @@
...
+): boolean {
+  if (getPartId(which) === id) return true;
+  visitUrl("place.php?whichplace=scrapheap&action=sh_configure");
+  visitUrl(`choice.php?whichchoice=1445&show=${which}`);
```
horrible-little-slime, line 111: Can't you just `runChoice` for a lot of these?

src/challengePaths/2021/YouRobot.ts
gausie: Oh I think I misunderstood

PR #637, opened 2024-08-09
Export gausie's magic tuple function

horrible-little-slime:


gausie (approved):
I guess I never exported because it seems like it'll eventually be a language feature.

PR #638, opened 2024-08-24
deprecate `flat` utility function; use `flat` method in its stead

horrible-little-slime:
Gausie, I know you have some code for polyfilling `matchAll`, and we can remove that too I think. Right, @migleyc?

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -863,8 +863,7 @@ export const holidayWanderers = new Map<string, Monster[]>([
 export function getTodaysHolidayWanderers(): Monster[] {
   return holiday()
     .split("/")
-    .map((holiday) => holidayWanderers.get(holiday) ?? [])
-    .flat();
+    .flatMap((holiday) => holidayWanderers.get(holiday) ?? []);
 }
 
 /**
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -1220,12 +1220,10 @@ export function setCombatFlags(
   return visitUrl(
     `account.php?${
       ([
-        flags
-          .map(({ flag, value }) => [
-            `actions[]=flag_${flag}`,
-            `flag_${flag}=${Number(value)}`,
-          ])
-          .flat(),
+        ...flags.flatMap(({ flag, value }) => [
+          `actions[]=flag_${flag}`,
+          `flag_${flag}=${Number(value)}`,
+        ]),
         "action=Update",
         "am=1",
         "ajax=1",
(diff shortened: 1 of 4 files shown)


midgleyc:
No, `matchAll` isn't supported (though the non-Symbol version probably isn't too hard to PR).

I think you'll want to update https://github.com/loathers/libram/blob/main/babel.config.cjs as well for the compiled version (I assume this is used?)

src/lib.ts
```diff
@@ -1221,12 +1220,12 @@ export function setCombatFlags(
   return visitUrl(
     `account.php?${
       ([
-        ...flat(
-          flags.map(({ flag, value }) => [
+        flags
```
midgleyc, line 1223: `...` has gone missing, think it's still needed?

Also I think you can `flatMap` here?

src/lib.ts
```diff
@@ -1221,12 +1220,10 @@ export function setCombatFlags(
   return visitUrl(
     `account.php?${
       ([
-        ...flat(
-          flags.map(({ flag, value }) => [
-            `actions[]=flag_${flag}`,
-            `flag_${flag}=${Number(value)}`,
-          ]),
-        ),
+        ...flags.flatMap(({ flag, value }) => [
```
midgleyc, line 1223: now that I read this in more detail this looks like another `[...x]` that can be simplified to `x`.

src/lib.ts
```diff
@@ -861,11 +861,10 @@ export const holidayWanderers = new Map<string, Monster[]>([
...
 export function getTodaysHolidayWanderers(): Monster[] {
-  return flat(
-    holiday()
-      .split("/")
-      .map((holiday) => holidayWanderers.get(holiday) ?? []),
-  );
+  return holiday()
+    .split("/")
+    .map((holiday) => holidayWanderers.get(holiday) ?? [])
+    .flat();
```
midgleyc, line 867: could flatMap

src/resources/2023/CursedMonkeyPaw.ts
```diff
@@ -48,24 +48,23 @@ export type WishableItemsFilters = Partial<{
...
+              .filter(
+                ({ type, rate, drop }) =>
+                  !drop.quest &&
+                  (type !== "c" || rate >= 1) && // Remove random roll drops
+                  (filters.drop?.({ type, rate, drop }) ?? true),
+              )
+              .map(({ drop }) => drop),
+          ),
+      )
+      .flat(2),
```
midgleyc, line 67: could flatMap

PR #639, opened 2024-08-25
spice up & export getScalingCap and getScalingRate

horrible-little-slime:


src/lib.ts
```diff
@@ -1282,3 +1283,46 @@ export function extractItems(text: string): Map<Item, number> {
...
+    const current = cache.get(monster);
+    if (current !== undefined) return monsterEval(current);
+
+    const result = pattern.exec(monster.attributes);
+    if (!result) {
+      cache.set(monster, "0");
+      return 0;
+    }
+    cache.set(monster, result[1]);
+    return monsterEval(result[1]);
```
Rinn, line 1301: You can simplify this with optional chaining and nullish coalescing:
```suggestion
    const result = pattern.exec(monster.attributes)?.[1] ?? "0";
    cache.set(monster, result);
    return monsterEval(result);
```

Rinn (approved):

PR #642, opened 2024-09-03
Value JOAT as a 1.5x fairy

Ignose:


src/lib.ts
```diff
@@ -837,6 +837,10 @@ export function findFairyMultiplier(familiar: Familiar): number {
   if (familiar === $familiar`Reanimated Reanimator`) return 0;
   const itemBonus = numericModifier(familiar, "Item Drop", 1, $item.none);
   if (itemBonus === 0) return 0;
+  if (familiar === $familiar`Jill-of-All-Trades`)
+    return (
+      Math.pow(Math.sqrt(itemBonus + 55 / 4 + 3) - Math.sqrt(55) / 2, 2) * 1.5
+    );
```
horrible-little-slime, line 843:
```suggestion
    return 1.5;
```

Ignose: I'm confused; normally we return `return Math.pow(Math.sqrt(itemBonus + 55 / 4 + 3) - Math.sqrt(55) / 2, 2);` - why would we only want to return 1.5?

gausie: It is a little counterintuitive but the big bit of maths there is actually working out the multiplier based on the Item Drop at 1lb. Perhaps you, as I did at first glance, thought it was something to do with calculating item drop from familiar weight. Probably accept this change and add a comment to the final return reflecting this clarification for good measure

gausie (approved):
Style note is that I like the comments to be on the line before the code. It keeps lines shorter and means I'm "prepared" for what I'm about to read, rather than confused and THEN comforted. Coding is weird. And maybe your first comment should sort of say something like "We just go ahead and assume if the player is asking the Fairy Multiplier of JOAT, they're gonna tune the LED candle appropriately"

Nothing to stop me approving though.

PR #643, opened 2024-09-05
Conspiracy Island

gausie:
I am trying to clear out my personal scripts folder (particularly ash scripts) without throwing away the actual work I did. This represents the libram-appropriate bits of my GausieConspiracy.ash file. I have some more code to give status updates on the bunker machine and to actually finish quests, but the former is just mostly useless and the latter is better constructed from these functions.

For posterity/inspiration here are my quest completion functions. These are essentially the "tasks" to be run in the quest location. If there is no function, you basically just run in that location until the quest flag says finished.

```java
void complete_eve_quest( ConspiracyQuest quest )
{
    quest.loc.run_until( "beaten_eve" );

    while ( get_property( quest.prop ) != quest.complete )
    {
        quest.loc.adventure( 1 );
    }
}

void complete_serum_quest( ConspiracyQuest quest )
{
    if ( retrieve_item( 5, $item[experimental serum P-00] ) )
    {
        return;
    }

    while ( get_property( quest.prop ) != quest.complete )
    {
        quest.loc.adventure( 1 );
    }
}

void complete_smokes_quest( ConspiracyQuest quest )
{
    maximize( `item`, false );

    monster smoke = $monster[smoke monster];

    set_property( "autoOlfact", `monster {smoke}` );

    while ( get_property( quest.prop ) != quest.complete )
    {
        set_property( "autoPutty", `monster {smoke}` );

        quest.loc.adventure( 1 );

        smoke.fight_copy();
    }
}
```

### DIFF
--- a/src/resources/2014/ConspiracyIsland.ts
+++ b/src/resources/2014/ConspiracyIsland.ts
@@ -1,0 +1,188 @@
+import { visitUrl, runChoice, Location, handlingChoice } from "kolmafia";
+import { questStep } from "../../lib.js";
+import { Requirement } from "../../maximize.js";
+import { get } from "../../property.js";
+import { $location, $item } from "../../template-string.js";
+
+class ConspiracyQuest {
+  prop: string;
+  reward: number;
+  complete: number;
+  /** Location in which quest is completed */
+  location: Location;
+  /** Requirements to complete quest */
+  requirements: Requirement;
+
+  constructor(
+    prop: string,
+    reward: number,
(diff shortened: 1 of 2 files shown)


Rinn:
nice. mitght be worth adding a few helper functions to get if the shops are unlocked, and to get the location & keycard needed to unlock them?

Rinn (approved):

gausie:
In future, absolutely.

PR #651, opened 2024-09-12
Fix bad banisher in pref

gausie:
- **Skill.get throws an error, does not return .none**


src/lib.ts
```diff
@@ -499,11 +499,9 @@ export function getBanishedMonsters(): Map<Item | Skill, Monster> {
         null,
       ].includes(banisherItem)
     ) {
-      if (Skill.get(banisher) === $skill.none) {
-        break;
-      } else {
-        result.set(Skill.get(banisher), Monster.get(foe));
-      }
+      const skill = $skill.get(banisher);
+      if (!skill) break;
```
horrible-little-slime, line 503:
```suggestion
      if (!skill) continue;
```

horrible-little-slime (changes requested):

src/lib.ts
gausie: Yeah fair enough. This is a behavioural change but is desired probably. I think it was originally break to deal with an indivisible-by-3 number of elements or something

PR #654, opened 2024-09-13
noneToNull cover all classes

gausie:


src/lib.ts
```diff
@@ -550,19 +551,8 @@ export function canUse(item: Item): boolean {
  * @returns The thing specified or `null`
  */
 export function noneToNull<T>(thing: T): T | null {
```
horrible-little-slime, line 553:
```suggestion
export function noneToNull<T extends MafiaClass>(thing: T): T | null {
```

horrible-little-slime (changes requested):

src/lib.ts
gausie: Not descriptively true, but perfectly prescriptively true yes.

PR #664, opened 2024-09-23
new BALLS stuff

horrible-little-slime:


src/Macro.test.ts
```diff
@@ -10,8 +10,13 @@ import {
   $skill,
   $stat,
 } from "./template-string.js";
+import { before } from "node:test";
```
gausie, line 13: Import from vitest

src/Macro.test.ts
```diff
@@ -10,8 +10,13 @@ import {
...
 
 describe(Macro, () => {
+  before(() => {
+    Object.assign($skill`mock skill`, { combat: true });
```
gausie, line 17: You probably need to mock Skill.get here

PR #669, opened 2024-10-12
perf: Memoize getFoldGroup.

phulin:


horrible-little-slime (approved):
Had to manually check if items are in their own fold groups (they are) so this seems fine.

PR #672, opened 2024-10-14
Bridge skip for Libram

Ignose:


Ignose:
I am working on the belief that https://wiki.kolmafia.us/index.php/Quest_Tracking_Preferences is correctly documented, which is what LoopRobot relies on so it should be.

src/resources/2024/BatWings.ts
```diff
@@ -38,3 +39,19 @@ export function cauldronsRemaining(): number {
...
+/**
+ * Attempts to do bridge skip
+ * @returns whether we were successfully able to skip the bridge
+ */
+export function bridgeSkip(): boolean {
+  if (get("chasmBridgeProgress") < 25 || get("questL09Topping") !== "started") {
+    return false;
+  }
+  visitUrl(
+    `place.php?whichplace=orc_chasm&action=bridge${get("chasmBridgeProgress")}`,
```
ajcoppa, line 52: Trying to build the bridge here doesn't really make sense given that the first clause of the if statement blocks getting here if they haven't built enough already. We should decide whether we want to require callers to handle their own bridge progress or if libram is responsible for handling that for them and be consistent.

(I'm in support of libram helping build the bridge too, but that's more up to @horrible-little-slime.)

src/resources/2024/BatWings.ts
```diff
@@ -38,3 +39,19 @@ export function cauldronsRemaining(): number {
 export function flapChance(flaps = get("_batWingsFreeFights")): number {
   return flaps < 5 ? 1 / (2 + flaps) : 0;
 }
+
+/**
+ * Attempts to do bridge skip
+ * @returns whether we were successfully able to skip the bridge
+ */
+export function bridgeSkip(): boolean {
```
ajcoppa, line 47: You need wings equipped in order to make the skip available. Should libram handle that too or do we want callers to handle that?

horrible-little-slime (commented):
I agree with Zinc that this needs some work. I can see two incarnations of how this might work: an extremely atomic version that _just_ checks if you're specifically in position to jump the bridge and does it, and a larger, spicier version that includes everything. The former should look something like this:
- Check bridge progress. If it's < 25, return false. If it's >= 30, return true
- Check to see if you have bat wings. If you don't, return false
- Do the jump, and return whether you have access to the highlands at the end.
That version doesn't include attempting to build up the bridge any further than you have it, and unambiguously belongs as a bat wings function imo. 

The latter is a fuller "attempt to build the bridge" function, and it may belong in `lib` rather than in `BatWings` because it's more general
- check to see if the quest is in an appropriate state (that is, you're on the "building the bridge" step of the L9 quest regardless of bridge status). If the quest is unstarted, return false; if the bridge is built, return true. The final return value should always just be "can we access the highlands"
- attempt to build the bridge
- if the bridge is complete, return true
- if you don't have bat wings, return false
- if the bridge is < 25 progress, return false
- do the jump, and return success status

I think realistically the answer is both: add a `jumpBridge` function to batwings that you use for the final few steps of a general `bridgeChasm` function. There's a bit of a circular imports problem here--`bridgeChasm`, as it currently stands, belongs in lib, and lib shouldn't import anything outside of utils. I think for now implement the atomic `jumpBridge` function, and we can talk about adding a `quests.ts` file (or, like, an unrestricted/standard module in our challenge paths directory) for functions like `bridgechasm`

horrible-little-slime:
Oh, and use `questStep` for quest prefs

src/resources/2024/BatWings.ts
```diff
@@ -38,3 +39,26 @@ export function cauldronsRemaining(): number {
...
```
horrible-little-slime, line 47:
```suggestion
export function jumpBridge(): boolean {
```

PR #673, opened 2024-10-14
Fix default behavior for get of known strings properties.

phulin:


src/property.ts
```diff
@@ -196,7 +196,7 @@ export function get(property: string, _default?: unknown): unknown {
   } else if (isPhylumProperty(property)) {
     return getPhylum(property, _default as Phylum | undefined);
   } else if (isStringProperty(property)) {
-    return value;
+    return value === "" && _default !== undefined ? _default : value;
```
horrible-little-slime, line 199:
```suggestion
    return value || _default ?? "";
```

horrible-little-slime (approved):
fine!

PR #679, opened 2024-11-04
Add Stillsuit features

Ignose:
This has yet to be tested, but offers the ability to pop stillsuit as an item% buff based on regex of the distillate, among other features.

### DIFF
--- a/src/resources/2022/Stillsuit.ts
+++ b/src/resources/2022/Stillsuit.ts
@@ -39,14 +39,23 @@ export function drinkDistillate(): boolean {
  * @param modifier determines what modifier to check stillsuit buffs against
  * @returns the modifier value for the given modifier
  */
-export function distillateModifier(modifier: string): number {
+export function distillateModifier(modifier: NumericModifier): number {
   // Retrieve the current distillate modifiers as a string
   const distillateMods = get("currentDistillateMods");
 
+  const experienceMap: Record<string, string> = {
+    "Muscle Experience": "Experience (Muscle)",
+    "Mysticality Experience": "Experience (Mysticality)",
+    "Moxie Experience": "Experience (Moxie)",
+  };
+
+  // Adjust the modifier if it is one of the special cases
+  const adjustedModifier = experienceMap[modifier] ?? modifier;
+
   // Construct a regex pattern to match the modifier and capture the numeric value
-  const regex = new RegExp(`${modifier}: \\+?(-?\\d+)`);
+  const regex = new RegExp(`${adjustedModifier}: \\+?(-?\\d+)`);
   const match = distillateMods.match(regex);
 
   // If a match is found, parse and return the captured number; otherwise, return 0
-  return match ? parseInt(match[1], 10) : 0;
+  return match ? Number(match[1]) : 0;
 }
--- a/src/resources/2022/Stillsuit.ts
+++ b/src/resources/2022/Stillsuit.ts
@@ -22,7 +22,7 @@ export function have(): boolean {
 export function distillateAdventures(): number {
   if (!have()) return 0;
   const sweat = get("familiarSweat");
-  return Math.round(sweat ^ 0.4);
+  return Math.round(sweat ** 0.4);
 }
 
 /**
@@ -31,8 +31,7 @@ export function distillateAdventures(): number {
  */
 export function drinkDistillate(): boolean {
   if (!have() || get("familiarSweat") <= 0) return false;
-  cliExecute("drink stillsuit distillate");
-  return true;
+  return cliExecute("drink stillsuit distillate");
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/resources/2022/Stillsuit.ts
```diff
@@ -0,0 +1,79 @@
...
+  return haveItem($item`tiny stillsuit`);
+}
+
+/**
+ * @returns the expected adventures from drinking stillsuit distillate
+ */
+export function distillateAdventures(): number {
+  if (!have()) return 0;
+  const sweat = get("familiarSweat");
+  return Math.round(sweat ^ 0.4);
```
horrible-little-slime, line 20: [This does not do what you think it does](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR)

instead use ** or Math.pow

src/resources/2022/Stillsuit.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * Drinks stillsuit distillate
+ * @returns whether distillate was successfully drunk
+ */
+export function drinkDistillate(): boolean {
+  if (!have() || get("familiarSweat") <= 0) return false;
+  cliExecute("drink stillsuit distillate");
+  return true;
```
horrible-little-slime, line 30:
```suggestion
  return cliExecute("drink stillsuit distillate");
```

src/resources/2022/Stillsuit.ts
```diff
@@ -0,0 +1,79 @@
...
+  "Sleaze Spell Damage": "Sleaze Spell Damage",
+  "Hot Spell Damage": "Hot Spell Damage",
+  "Cold Spell Damage": "Cold Spell Damage",
+  "Stench Spell Damage": "Stench Spell Damage",
+  "Spooky Spell Damage": "Spooky Spell Damage",
+};
+
+// Regex to capture both "+X [modifier]" and "[modifier] +X" or "[modifier]: X"
+const regex =
+  /(?:\+(\d+)%?\s*(Muscle Stats Per Fight|Mysticality Stats Per Fight|Moxie Stats Per Fight|Item Drops from Monsters|Food Drops from Monsters|Combat Initiative|Spooky Damage|Hot Damage|Cold Damage|Stench Damage|Sleaze Damage|Weapon Damage|Damage Reduction|Sleaze Spell Damage|Hot Spell Damage|Cold Spell Damage|Stench Spell Damage|Spooky Spell Damage)|(?:Damage Reduction|Weapon Damage):?\s+(\d+))/g;
```
horrible-little-slime, line 56: This cannot possibly be the best way to do this

src/resources/2022/Stillsuit.ts
```diff
@@ -0,0 +1,79 @@
...
+      const value = parseInt(match[1] ?? match[3], 10); // Supports both formats (before or after)
+      const modifierText = match[2] || match[4];
+
+      // Map to standardized modifier names
+      const standardizedModifier = toModifier(modifierMap[modifierText]);
+
+      // Return tuple if mapping exists
+      return standardizedModifier ? [standardizedModifier, value] : null;
+    })
+    .filter((result): result is [Modifier, number] => result !== null); // Adjusted type predicate
```
horrible-little-slime, line 76: Instead of doing a single massive RegExp to handle this, I would either iterate over the set of available modifiers and populate an object that way, or just `split` `nextDistillateMods` by `, `, then `split` its constituent elements by `: ` and then map some of the weirder-named modifiers to their correct NumericModifier names in mafia, returning a `Partial<Record<NumericModifier, number>>`

horrible-little-slime (changes requested):

src/resources/2022/Stillsuit.ts
```diff
@@ -1,7 +1,12 @@
-import { cliExecute, Modifier, toModifier, visitUrl } from "kolmafia";
+import { cliExecute, visitUrl } from "kolmafia";
 import { have as haveItem } from "../../lib.js";
 import { get } from "../../property.js";
 import { $item } from "../../template-string.js";
+import {
+  arrayContains,
+  NumericModifier,
+  numericModifiers,
+} from "../../index.js";
```
horrible-little-slime, line 9: never import from index.js

src/resources/2022/Stillsuit.ts
```diff
@@ -30,50 +35,27 @@ export function drinkDistillate(): boolean {
...
+  visitUrl("inventory.php?action=distill&pwd"); // Update the mafia pref
+  const distillateModsString = get("currentDistillateMods");
+
+  const distillateMods: Record<NumericModifier, number> = distillateModsString
+    .split(", ")
+    .reduce(
+      (acc, modString) => {
+        const [key, value] = modString.split(": ");
+        if (key && value && arrayContains(key, numericModifiers)) {
+          acc[key as NumericModifier] = parseInt(value.replace("+", ""), 10);
```
horrible-little-slime, line 53: You probably don't need to cast `key` if you use `arrayContains`?

also, imo just use `Number` over `parseInt`

src/resources/2022/Stillsuit.ts
```diff
@@ -30,50 +35,27 @@ export function drinkDistillate(): boolean {
...
+        }
+        return acc;
+      },
+      {} as Record<NumericModifier, number>,
+    );
+
+  return distillateMods[modifier] || 0;
```
horrible-little-slime, line 60: If the function exists only to return a single modifier, it actually does make sense to just use a regexp. Because you can just hunt down that specific one and look for `${modifier}: +?(\d+), `

horrible-little-slime (approved):

PR #684, opened 2024-11-13
Stillsuit-ii

horrible-little-slime:


src/resources/2022/Stillsuit.ts
```diff
@@ -56,3 +57,93 @@ export function distillateModifier(modifier: NumericModifier): number {
...
+  familiar: Familiar,
+): Partial<Record<NumericModifier, number>> {
+  const tags = getFamiliarTags(familiar);
+  return tags
+    .filter(isStillsuitTag)
+    .reduce<Partial<Record<NumericModifier, number>>>(
+      (acc, tag) => ({
+        ...acc,
+        [MODIFIER_TAGS[tag]]:
+          ((acc[MODIFIER_TAGS[tag]] ?? 0) + 1) / tags.length,
```
gausie, line 133: Technicality but is `tags.length` right here for `pokefam`-tag-having familiars?

Ignose: Could you just do the filter earlier to solve this?

`const tags = getFamiliarTags(familiar).filter(isStillsuitTag);`

horrible-little-slime:
There's nothing to solve; this is accurate to how the stillsuit works with
pokefam ult tags

gausie (approved):

PR #692, opened 2024-11-22
Add Comma Chameleon functionality

Ignose:


src/resources/2006/CommaChameleon.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Transforms Comma Chameleon into the familiar of choice
+ * @param fam determines what to transform into
+ * @returns Whether Comma has been successfully transformed
+ */
+export function transform(fam: Familiar): boolean {
+  let equipment = null;
+  equipment = familiarEquipment(fam);
+  if (!equipment) return false;
+  if (!_have(equipment)) return false;
```
horrible-little-slime, line 26:
```suggestion
  const equipment = familiarEquipment(fam);
  if (equipment !== $item.none) return false;
  if (!have_(equipment)) return false;
```

In general it's good to try to avoid using `let`; here, there's literally no utility to it given that you assign the variable and then _immediately_ reassign it.

`$item.none` isn't falsy, and `familiarEquipment` returns it in the null case (this will be true for virtually all ash functions)

src/resources/2006/CommaChameleon.ts
```diff
@@ -0,0 +1,31 @@
+import { $familiar } from "../../template-string.js";
+import { have as _have } from "../../lib.js";
```
horrible-little-slime, line 2:
```suggestion
import { have as have_ } from "../../lib.js";
```

putting a `_` in front is for when something is intended to be private; putting a `_` in the back is used for name collisions

src/resources/2006/CommaChameleon.ts
```diff
@@ -0,0 +1,31 @@
...
+const familiar = $familiar`Comma Chameleon`;
+
+/**
+ * Determines whether the player has the Comma Chameleon in their
+ * terrarium
+ *
+ * @returns Whether the player has a Comma Chameleon
+ */
+export function have(): boolean {
+  return _have(familiar);
```
horrible-little-slime, line 14:
```suggestion
  return have_(familiar);
```

src/resources/2006/CommaChameleon.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+/**
+ * Transforms Comma Chameleon into the familiar of choice
+ * @param fam determines what to transform into
+ * @returns Whether Comma has been successfully transformed
+ */
```
horrible-little-slime, line 21: clarify that it doesn't do anything to acquire the appropriate item

horrible-little-slime (commented):
add a function that tells us what familiar it's currently acting as, and gate `transform` behind a check for that as well (assuming that works)

src/resources/2006/CommaChameleon.ts
Ignose: Odd. I copied from the Bander code.

PR #695, opened 2024-12-01
fix: [[alpha]] is not a valid character class

midgleyc:


src/resources/2017/Spacegate.ts
```diff
@@ -179,7 +179,7 @@ export function dial(address: string): boolean {
     //cannot dial if we already have or don't own it.
     return false;
   }
-  if (!address.match(`^[[alpha]]+$`) || address.length !== 7) {
+  if (!address.match(`^[A-Za-z]+$`) || address.length !== 7) {
```
horrible-little-slime, line 182:
```suggestion
  if (!address.match(/^[A-Za-z]+$/) || address.length !== 7) {
```

might as well use a real regexp while we're here

PR #703, opened 2024-12-18
create & export `getPrediction` from CrystalBall

horrible-little-slime:


src/resources/2021/CrystalBall.ts
```diff
@@ -22,20 +22,28 @@ export const orb = Item.get("miniature crystal ball");
...
+/**
+ * @returns A map keyed by locations and whose values are monsters, representing all active orb predictions
+ */
+export function getPrediction(): Map<Location, Monster> {
+  return new Map(
+    get("crystalBallPredictions")
+      .split("|")
+      .filter(Boolean)
+      .map((element) => element.split(":") as [string, string, string])
+      .filter((tuple) => tuple.length === 3)
```
gausie, line 35: It might make sense to make a utility that parses multi dimensional arrays encoded as strings like this into arrays of tuples, as it's a common mafia pattern. And might as well export that utility.

gausie (approved):

PR #734, opened 2025-02-07
npcPrice for getAcquirePrice

Shiverwarp:
I don't know how much more we want to really mess with this, it feels like it's difficult to get things to work as we wish


horrible-little-slime (approved):
I guess?

PR #750, opened 2025-03-04
Add initial maps, much work to be done

Ignose:


src/resources/2025/LepreCondo.ts
```diff
@@ -0,0 +1,315 @@
...
+ */
+export function have(): boolean {
+  return have_($item`Leprecondo`);
+}
+
+/**
+ * @returns How many more times you can rearrange today
+ */
+export function rearrangeUses(): number {
+  return get("_leprecondoRearrangements");
```
spaghetti-squash, line 17: don't you need 3 - this?

src/resources/2025/LepreCondo.ts
```diff
@@ -0,0 +1,315 @@
...
+}
+
+function parseFurniture(str: string): Furniture[] {
+  return str
+    .split(",")
+    .map(Number)
```
spaghetti-squash, line 23: no need for this to be two separate maps, you can do `.map((index) => itemsMap[Number(index)])`

src/resources/2025/LepreCondo.ts
```diff
@@ -0,0 +1,315 @@
...
+
+type Need =
+  | "exercise"
+  | "mental stimulation"
+  | "dumb entertainment"
+  | "food"
+  | "booze"
+  | "sleep";
+
+const itemsMap: Record<
```
spaghetti-squash, line 116: agree with gausie that this should just be an array

src/resources/2025/LepreCondo.ts
```diff
@@ -0,0 +1,315 @@
...
+      discoveredFurniture.some((d) => d.number === furniture?.number),
+    )
+  ) {
+    return false;
+  }
+
+  const furnitureNumbers = [f1, f2, f3, f4].map((f) => f?.number ?? 0); // Default to 0 if undefined
+  directlyUse($item`Leprecondo`);
+  visitUrl(
+    `choice.php?pwd&option=1&whichchoice=1556&r0=${furnitureNumbers[0]}&r1=${furnitureNumbers[1]}&r2=${furnitureNumbers[2]}&r3=${furnitureNumbers[3]}`,
```
spaghetti-squash, line 72: just use runChoice

src/resources/2025/LepreCondo.ts
spaghetti-squash: Also given that the elements of `itemsMap` aren't themselves `Furniture` by the typedef you lay down below, this won't work great

src/resources/2025/LepreCondo.ts
Ignose: runChoice didn't work for me in testing. Probably an error on my part.

PR #779, opened 2025-04-04
Offer shield mood skills/effects

Ignose:


src/mood.ts
```diff
@@ -303,6 +310,70 @@ class AsdonMoodElement extends MoodElement {
...
+    this.skill = skill;
+  }
+
+  mpCostPerTurn(): number {
+    const turns = turnsPerCast(this.skill);
+    return turns > 0 ? mpCost(this.skill) / turns : 0;
+  }
+
+  turnIncrement(): number {
+    return turnsPerCast(this.skill);
```
spaghetti-squash, line 331: I don't think we can rely on turnsPerCast here, we might need to add it to the map

src/mood.ts
```diff
@@ -303,6 +310,70 @@ class AsdonMoodElement extends MoodElement {
...
+    let oldRemainingCasts = -1;
+    let remainingCasts = Math.ceil(
+      (ensureTurns - initialTurns) / turnsPerCast(this.skill),
+    );
+
+    while (remainingCasts > 0 && oldRemainingCasts !== remainingCasts) {
+      equip($item`April Shower Thoughts shield`);
+
+      let maxCasts;
+      if (hpCost(this.skill) > 0) {
```
spaghetti-squash, line 349: Not relevant for shield

src/mood.ts
```diff
@@ -303,6 +310,70 @@ class AsdonMoodElement extends MoodElement {
...
+          const bestMp = Math.min(remainingCasts * cost, myMaxmp());
+          mood.moreMp(bestMp);
+          maxCasts = Math.floor(Math.min(mood.availableMp(), myMp()) / cost);
+        }
+      }
+
+      const casts = clamp(remainingCasts, 0, Math.min(100, maxCasts));
+      useSkill(casts, this.skill);
+
+      unequip($item`April Shower Thoughts shield`);
```
spaghetti-squash, line 365: Rather than using unequip we should restore your previous offhand

spaghetti-squash: Also for regular skills we need to remove shield for empathy

spaghetti-squash (changes requested):

spaghetti-squash:
It'd be nice if there was a way to force all april skills to be done in a row with the shield up, so we aren't taking it on and off for each skill (this is still WAY better than for each cast). @phulin is the resident expert on mood

src/mood.ts
```diff
@@ -24,12 +27,27 @@ import {
...
 import { get } from "./property.js";
 import { AsdonMartin } from "./resources/index.js";
-import { $item, $skill } from "./template-string.js";
+import { $effect, $item, $skill, $slot } from "./template-string.js";
 import { clamp, sum } from "./utils.js";
 
+const shieldEffects = new Map<Effect, { skill: Skill; turnsPerCast: number }>([
+  [
+    $effect`Thoughtful Empathy`,
+    { skill: $skill`Empathy of the Newt`, turnsPerCast: 5 },
```
spaghetti-squash, line 39: This one does actually obey turns per cast, because it's functionally treated as a turtle totem spell

src/mood.ts
```diff
@@ -24,12 +27,27 @@ import {
...
+  ],
+  [
+    $effect`Lubricating Sauce`,
+    { skill: $skill`Sauce Contemplation`, turnsPerCast: 5 },
```
spaghetti-squash, line 43: I misremembered these as giving fewer adventures than the regular spell does; looks like we _can_ rely on turnsPerCast after all, just as sort-of-a-coincidence. Sorry king

src/mood.ts
```diff
@@ -136,9 +154,17 @@ class SkillMoodElement extends MoodElement {
   execute(mood: Mood, ensureTurns: number): boolean {
     const effect = toEffect(this.skill);
     const initialTurns = haveEffect(effect);
+    let showerShieldUnequipped = false;
```
spaghetti-squash, line 157: instead of storing this in a `let` variable, use a `const` to store the slot it's equipped in:
```typescript
  const aprilShowerShieldSlot = Slot.all().find((s) => equippedItem(s) === $item`April Shower Thoughts shield`);
```
```
if (this.skill === $skill`Empathy of the Newt` && aprilShowerShieldSlot) { 
  unequip(aprilShowerShieldSlot);
}
```
```
if (aprilShowerShieldSlot) equip(aprilShowerShieldSlot, $item`April Shower Thoughts shield`);
```

Ignose:
A @phulin review would be lit; no rush though.

PR #786, opened 2025-04-11
update modifier list; handle Raw Combat Rate in libram

spaghetti-squash:
It's not 100% clear how much of this is needed. I'm also not confident I love the handling of multistrings in `getModifier` given that most of the time an item applies only one effect.


spaghetti-squash:
Waiting for https://github.com/kolmafia/kolmafia/pull/2886 before merging
This'll involve cutting a new major version because it'll change the behavior of `getModifiers`

PR #792, opened 2025-04-30
don't require little firkins to smash barrels

spaghetti-squash:


docrostov (approved):
after discussing in discord i trust that this is correct and proper even though the dark magick is still unattainable for my hedonistic eyes

PR #797, opened 2025-05-16
Quest prefs

spaghetti-squash:


tools/parseDefaultProperties.ts
```diff
@@ -140,6 +140,15 @@ export function isItemProperty(property: string): boolean {
   );
 }
 
+const underscoredQuests = ["_questPartyFair", "_questPirateRealm"]; // Most prefs that start with _quest are not quests
+/**
+ * @param property Property name
+ * @returns Whether the supplied property should be coerced into a QuestStep
+ */
+export function isQuestProperty(property: string): boolean {
+  return property.startsWith("quest") || underscoredQuests.includes(property);
```
gausie, line 149: Why not do || starts with _quest?

spaghetti-squash:
Because of all of the NEP quest prefs that start with `_quest` that aren't
the quest

PR #803, opened 2025-06-06
Initial Beret Support

Ignose:
Includes have and an optimizer function.

Optimizer function could be optimized, probably.

### DIFF
--- a/src/resources/2025/PrismaticBeret.ts
+++ b/src/resources/2025/PrismaticBeret.ts
@@ -271,8 +271,9 @@ function findOutfit(power: number, buyItems: boolean, hammerTime: boolean) {
 
   for (const slot of relevantSlots) {
     const item = outfit[slot];
+    if (item === Item.none) continue;
     if (have_(item)) continue;
-    if (!buy(item) || item === Item.none) {
+    if (!buy(item)) {
       logger.debug(`Failed to purchase ${item}`);
       return null;
     }


src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,122 @@
...
+        pants.flatMap((pant) => shirts.flatMap((shirt) => hat + pant + shirt)),
+      ),
+    ),
+  ];
+}
+
+function scoreBusk(
+  effects: Effect[],
+  weightedModifiers: [Modifier, number][],
+  uselessEffects: Effect[],
```
spaghetti-squash, line 67:
```suggestion
  uselessEffects: Set<Effect>,
```

src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,122 @@
...
+): number {
+  const usefulEffects = effects.filter((ef) => !uselessEffects.includes(ef));
```
spaghetti-squash, line 69:
```suggestion
  const usefulEffects = effects.filter((ef) => !uselessEffects.has(ef));
```

src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,122 @@
...
+ * @param uselessEffects accepts an input of Effect[] and values those effects at 0 for scoring
+ * @param buyitem boolean, determines if we should check shop available equipment; defaults to true
+ */
+export function findOptimalBusks(
+  weightedModifiers: [Modifier, number][],
+  start?: number,
+  uselessEffects: Effect[] = $effects``,
+  buyitem: boolean = true,
+): number {
+  const buskUses = clamp(toInt(get("_beretBuskingUses")), 0, 5);
```
spaghetti-squash, line 91:
```suggestion
```

src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * @returns the equipment power required to get the optimal buffs for a given busk
+ * @param weightedModifiers accepts an input of [Modifier, number][] to determine how to optimize for buffs and which to prioritize
+ * @param start accepts an input of number (0-5) to determine which busk to check; default behavior is to check the next available busk
+ * @param uselessEffects accepts an input of Effect[] and values those effects at 0 for scoring
+ * @param buyitem boolean, determines if we should check shop available equipment; defaults to true
+ */
+export function findOptimalBusks(
+  weightedModifiers: [Modifier, number][],
```
spaghetti-squash, line 86:
```suggestion
  weightedModifiers: Partial<Record<NumericModifier, number>>,
```

src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,122 @@
...
+  start?: number,
+  uselessEffects: Effect[] = $effects``,
```
spaghetti-squash, line 88:
```suggestion
  uselessEffects: Effect[] = [],
```

src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,122 @@
...
+  buyitem: boolean = true,
+): number {
+  const buskUses = clamp(toInt(get("_beretBuskingUses")), 0, 5);
+
+  const targetBuskIndex = start ?? buskUses;
```
spaghetti-squash, line 93:
```suggestion
```

src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,122 @@
...
+ * @param buyitem boolean, determines if we should check shop available equipment; defaults to true
+ */
+export function findOptimalBusks(
+  weightedModifiers: [Modifier, number][],
+  start?: number,
+  uselessEffects: Effect[] = $effects``,
+  buyitem: boolean = true,
+): number {
+  const buskUses = clamp(toInt(get("_beretBuskingUses")), 0, 5);
+
```
spaghetti-squash, line 92:
```suggestion
const uselessEffectSet = new Set(uselessEffects);
```

src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,273 @@
...
+  const shirtPowers = populateMap(useableShirts, power, false);
+
+  for (const [hatPower, hat] of hatPowers) {
+    for (const [pantPower, pant] of pantsPowers) {
+      if (pantPower + hatPower > power) continue;
+      for (const [shirtPower, shirt] of shirtPowers) {
+        if (hatPower + shirtPower + pantPower === power)
+          if (buyItem) {
+            [hat, pant, shirt].forEach((i) => {
+              if (!have_(i) && npcPrice(i) > 0) buy(i);
```
spaghetti-squash, line 217: Do we know how this should behave if we fail to make a purchase? should we return null? we probably shouldn't try to buy the subsequent items

Ignose: if (!outfit) return false;

Passing null should? make this return false, and checking myMeat before we buy should make sure we don't buy anything if we can't afford everything.

src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,286 @@
...
+    for (const [pantPower, pant] of pantsPowers) {
+      if (pantPower + hatPower > power) continue;
+      for (const [shirtPower, shirt] of shirtPowers) {
+        if (hatPower + shirtPower + pantPower === power)
+          if (buyItem) {
+            const equips = [hat, pant, shirt];
+            const equipPrice = equips
+              .filter((i) => !have_(i))
+              .map((i) => npcPrice(i))
+              .reduce((a, b) => a + b, 0);
```
spaghetti-squash, line 221:
```suggestion
            const equipPrice = sum(equips
              .filter((i) => !have_(i)),
              npcPrice);
```

src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,286 @@
...
+
+            if (myMeat() < equipPrice) {
+              return null;
+            }
+
+            equips.forEach((i) => {
```
spaghetti-squash, line 227: We should decide partial success behavior

src/resources/2025/Beret.ts
```diff
@@ -0,0 +1,286 @@
...
+}
+
+const populateMap = (arr: Item[], max: number, double: boolean) => {
+  const map = new Map<number, Item>();
+  for (const it of arr) {
+    const power = getPower(it) * (double ? 2 : 1);
+    if (power > max) continue;
+
+    const existing = map.get(power);
+    if (!existing || (!have_(it) && npcPrice(it) < npcPrice(existing))) {
```
spaghetti-squash, line 189: Does this mean we don't populate the map if you _do_ have it?

spaghetti-squash:
We need to:
- determine what to do if you fail to buy an npc item
- ensure that we pick the npcprice-cheapest outfit for executing this
- by extension, only buy npc items if we must
I think that probably means we do a goofy little `maxBy` but that's fine
i'll re-lint it now

spaghetti-squash:
This needs:
- to account for durations
- testing
- a function that plugs optimal outfit directly into cast

MrFizzyBubbs:
I don't fully understand all of this, but here are my thoughts:
- Why is the value multiplied by duration in `valueEffect`?
- It looks like hammertime isn't accounted for
- I would be useful to have a function that returns the current "busk power" of the player's equipment. In CS I hardcore an outfit and like to check that it has the power that I expect before I busk.

spaghetti-squash:
> I don't fully understand all of this, but here are my thoughts:
> 
>     * Why is the value multiplied by duration in `valueEffect`?
> 
>     * It looks like hammertime isn't accounted for
> 
>     * I would be useful to have a function that returns the current "busk power" of the player's equipment. In CS I hardcore an outfit and like to check that it has the power that I expect before I busk.

Duration is basically an arbitrary choice--I think it makes equal sense to include and exclude it as a factor for the array case. No matter what, you can replicate your desired behavior by just manually using a function.

I thought I read that hammertime was accounted for in get_power, but you're right that looking at mafia code it doesn't seem to. How would a good API account for this? I don't think it should purely be statefully.

PR #824, opened 2025-10-06
Add Blood Cubic Zirconia to Libram functions

Ignose:


src/resources/2025/BloodCubicZirconia.ts
```diff
@@ -0,0 +1,79 @@
...
+export function availableCasts(skill: Skill, statFloor: number): number {
+  if (!have()) return 0;
+  if (!BCZSKILLS.includes(skill)) return 0;
+
+  const stat = BCZCOSTS.get(skill);
+  if (!stat) return 0;
+
+  // const currentStat = myBasestat(stat);
+  const currentStat = myBasestat(stat);
+  const timesCast = skill.timescast ?? 0;
```
spaghetti-squash, line 64: is this nullish?

src/resources/2025/BloodCubicZirconia.ts
```diff
@@ -0,0 +1,79 @@
...
+]);
+
+/**
+ * @param skill The BCZ skill to check.
+ * @param statFloor Minimum base stat you want to keep.
+ * @returns The number of casts you can achieve of the selected skill before hitting the given stat floor.
+ */
+export function availableCasts(skill: Skill, statFloor: number): number {
+  if (!have()) return 0;
+  if (!BCZSKILLS.includes(skill)) return 0;
```
spaghetti-squash, line 57: this check is redundant with the subsequent one imo

src/resources/2025/BloodCubicZirconia.ts
```diff
@@ -0,0 +1,79 @@
...
+  [13, 1100000],
+  [14, 2300000],
+  [15, 3700000],
+  [16, 11000000],
+  [17, 23000000],
+  [18, 37000000],
+  [19, 110000000],
+  [20, 230000000],
+  [21, 370000000],
+]);
```
spaghetti-squash, line 35: I feel like this should just be a function instead of a hardcoded list

src/resources/2025/BloodCubicZirconia.ts
spaghetti-squash, line 1: longer-term I think we will need a like `castDownTo(skill: Skill, statMinimum: number): number` (or maybe just return a boolean? idk)

src/resources/2025/BloodCubicZirconia.ts
```diff
@@ -0,0 +1,79 @@
+import { myBasestat, Skill, Stat } from "kolmafia";
+import { $item, $skill, $skills, $stat } from "../../index.js";
```
spaghetti-squash, line 2:
```suggestion
import { $item, $skill, $skills, $stat } from "../../template-string.js";
```
never import from the index. vscode loves to do it tho

src/resources/2025/BloodCubicZirconia.ts
```diff
@@ -0,0 +1,79 @@
...
+  const subStatFloor = statFloor ** 2;
+
+  let casts = 0;
+  let remainingStat = currentStat;
+
+  for (let i = timesCast; i < BCZSKILLCOST.size; i++) {
+    const nextCost = BCZSKILLCOST.get(i);
+    if (nextCost === undefined) break;
+    if (remainingStat - nextCost < subStatFloor) break;
+    remainingStat -= nextCost;
```
spaghetti-squash, line 74: Which is more reliable: tracking the stats you're at mafia, or asking for mafia's current understanding of your exp?
(I don't know)

package.json
```diff
@@ -48,7 +48,7 @@
     "eslint-config-prettier": "^9.1.0",
     "eslint-import-resolver-typescript": "^3.6.3",
     "eslint-plugin-jsdoc": "^50.2.3",
-    "eslint-plugin-libram": "^0.4.28",
+    "eslint-plugin-libram": "^0.4.32",
```
spaghetti-squash, line 51: i'll bump this literally the second after I submit this review

PR #828, opened 2025-11-07
cast function for bcz

Shiverwarp:


src/resources/2025/BloodCubicZirconia.ts
```diff
@@ -110,11 +110,32 @@ export function availableCasts(skill: Skill, statFloor: number): number {
  */
 export function castDownTo(skill: Skill, statFloor: number): boolean {
   if (!have() || !COSTS.get(skill)) return false;
-  let casts = availableCasts(skill, statFloor);
-  while (casts) {
-    if (!useSkill(skill, casts)) return false;
-    casts = availableCasts(skill, statFloor);
-  }
+
+  if (!cast(skill, availableCasts(skill, statFloor))) return false;
```
spaghetti-squash: I don't really remember why it was written the old way, but how concerned are we about mafia getting misaligned on a casts pref? we could do it as a one-at-a-time cast loop that repeatedly checks `availableCasts`, maybe? I guess this is strictly an improvement

src/resources/2025/BloodCubicZirconia.ts
```diff
@@ -110,11 +110,32 @@ export function availableCasts(skill: Skill, statFloor: number): number {
...
+  if (!have() || !COSTS.get(skill) || count <= 0) return false;
+  cliExecute("checkpoint");
+  try {
+    equip($item`blood cubic zirconia`);
+    let casts = count;
+    while (casts > 0) {
+      if (!useSkill(skill)) {
+        return false;
+      }
+      --casts;
```
spaghetti-squash:
```suggestion
    for (let i = 0; i < count; i++) {
      if (!useSkill(skill)) {
        return false;
      }
```

PR #834, opened 2025-12-10
Expose a withMacro function that lets you carry out some arbitrary action with the given macro

gausie:


spaghetti-squash (changes requested):

gausie:
Can't remember why I wanted this I'll just close it. Even though I think it's cool

PR #864, opened 2026-06-06
Name the clan that player's are not whitelisted to

libraryaddict:
Naming the clan in the error would reduce the number of reports being made, and make it easier to debug where the issue is. This isn't the first time I wished it named the clan while helping a user figure out why a script broke.

### DIFF
--- a/src/Clan.ts
+++ b/src/Clan.ts
@@ -123,7 +123,7 @@ export class Clan {
         );
 
         if (!clan) {
-          throw new Error("Player is not whitelisted to clan");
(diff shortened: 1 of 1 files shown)


spaghetti-squash (commented):
Should probably use `clanName` rather than `clanIdOrName`, right?

libraryaddict:
That's just `      const clanName = clanIdOrName.toLowerCase();`

spaghetti-squash (approved):
Ahh, you're right, I didn't catch that this only happens in the branch where the id is a string.

PR #867, opened 2026-07-16
cup of 13s

spaghetti-squash:


src/resources/2026/CupOfThirteens.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (
+    ingredients.some(
+      (ingredient) => !ingredient.tradeable || !ingredient.discardable,
+    )
+  )
+    return false;
+  visitUrl(`inventory.php?action=cupof13s&pwd=${myHash()}`, false);
+  return runChoice(
+    1,
+    [1, 2, 3].map((slot) => `whichitem${slot}`).join("&"),
```
soolar: Shouldn't this need to be something like this?
```suggestion
    [1, 2, 3].map((slot) => `whichitem${slot}=${ingredients[slot-1]}`).join("&"),
```
(not sure if I have Tuple usage syntax correct there)
Otherwise I don't see how the ingredients are actually being passed in here.

spaghetti-squash: oh great shout

spaghetti-squash: Alright I went with something slightly different but core conceit

soolar (approved):
lgtm with the caveat that I haven't verified myself that that is the exact correct format for the `runChoice` call, so make sure you've tested it

PR #870, opened 2026-08-06
Stop stale price from failing adson fills.

tottington:
`fillTo` aborts the calling script with `Failed to fuel Asdon Martin.` when the fuel it picked cannot be bought at the price ceiling it computed.

**The ceiling is built from a price that can be stale.** `getBestFuels` asks for `PriceAge.TODAY`, but `mallPrice` serves a cached value whenever mafia considers it fresh, and a bulk `mallPrices("food")` / `mallPrices("booze")` refresh marks everything fresh at once. 

Those two together mean a couple of meat of upward drift puts every real listing above the limit. `buy` searches, matches nothing, and acquires nothing silently, since acquiring less than requested is normal.

**`fillTo` did not check whether the purchase worked.** It called `insertFuel(bestFuel, Math.min(itemAmount(bestFuel), count))`, which is `0` when nothing was bought and none is held. It also masks the failure entirely whenever enough of the item happens to be in inventory already.

Changes:
- Allow the ceiling 10% of headroom for price drift.
- Walk the remaining candidates when a purchase yields nothing, instead of converting zero. Throw only once every candidate has been tried, naming them.

### DIFF
--- a/src/resources/2017/AsdonMartin.ts
+++ b/src/resources/2017/AsdonMartin.ts
@@ -193,6 +193,8 @@ export function insertFuel(it: Item, quantity = 1): boolean {
   return result.includes("The display updates with a");
 }
 
+const PRICE_DRIFT_TOLERANCE = 1.1;
+
 /**
  * Fill your Asdon Martin to the given fuel level in the cheapest way possible
  *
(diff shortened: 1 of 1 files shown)


gausie:
Closed in favour of #872

PR #872, opened 2026-08-08
Fix AsdonMartin.fillTo being bad

Rinn:


spaghetti-squash:
Have you taken a look at https://github.com/loathers/libram/pull/870 ?

Rinn:
> Have you taken a look at #870 ?

I hadn't no, but looking now afaik the ceiling variable was wrong before and still is because it is the next item efficiency * adv range, but the variable is a mall price max which would never have worker. My version should accurately buy the current item until it is no longer as efficient as the next item in the list.

Rinn:
I talked to them about #870, so we can take mine in favor of #870 since mine fixes the price cap

PR #874, opened 2026-08-13
Add basic laughing stock predictor

Ignose:


src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,149 @@
...
+export function canPredict(): boolean {
+  return have() && daycount() <= 4;
+}
+
+/**
+ * @param turnHorizon defines how many turns to check for available drops; defaults to remaining adventures
+ * @returns The next predictable Portable Laughing Stock drop and how many
+ * combats until it occurs, or null if the next drop is not predictable.
+ */
+export function nextDrop(turnHorizon = myAdventures()): [Item, number] | null {
```
gausie: Name the elements in this tuple

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,149 @@
...
+): Map<number, Item> {
+  const results = new Map<number, Item>();
+
+  let pityCount = 0;
+  let pityThreshold = 10;
+  let fixedPtr = 0;
+
+  for (let fight = 0; fight < maxFights; fight++) {
+    const seed =
+      classId ** 3 + 84 * pathId + 123 * (daycount - 1) + 381 * fight;
```
gausie: pull out seed as a separate function, it's good practice.

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,149 @@
...
+      } else {
+        hasDrop = false;
+      }
+    } else {
+      hasDrop = phpMtRand(rng, 1, 50) === 1;
+    }
+
+    if (!hasDrop) continue;
+
+    const advP = phpMtRand(rng, 1, 30);
```
gausie: Can you add some comments that explain what this actually is? I am trying to work out how we can pull the contents of this for loop into a separate function.

Ignose: I have no idea what this actually is, I stole the code from VeeAr! The phpMtRand stuff to me is like alchemy.

gausie: Ok well this is a poorly named variable so find out what it is please.

src/resources/2026/LaughingStock.ts
Ignose: Updated

src/resources/2026/LaughingStock.ts
Ignose: Updated

src/resources/2026/LaughingStock.ts
Ignose: Okay, so it's a boolean that tells us whether or not we get a drop between next fight and the number of fights defined by maxFights. It maps what combats we'll actually expect a drop.

gausie:
Correction on my last comment — I'd written that test with `toMatchSnapshot`, which nothing else in this repo does. Rewrote it as a plain `toEqual` against an explicit list and dropped the `__snapshots__` directory, then force-pushed over my own commit. Same failure, just legible in the diff now.

The substance is unchanged: the expected drops are the pre-triangular-n ones, and the deterministic half disagrees.

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,144 @@
...
+  Rng,
+} from "kolmafia";
+import { $item, $items } from "../../template-string";
+import { have as have_ } from "../../lib.js";
+import { get } from "../../property";
+import { clamp } from "../../utils";
+
+const basicFruit = $items`orange, grapefruit, grapes, lemon, lime, papaya, cranberries, strawberry, cherry, kumquat, tangerine, raspberry, kiwi, blackberry, banana, cactus fruit, plum, pear, peach`;
+
+const advFruit = $items`classic banana, antique watermelon, quince`;
```
spaghetti-squash:
```suggestion
const ADVANCED_FRUIT = $items`classic banana, antique watermelon, quince`;
```

characters are free, no need to scrimp and save

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,144 @@
...
+  phpMtRand,
+  phpSeed,
+  Rng,
+} from "kolmafia";
+import { $item, $items } from "../../template-string";
+import { have as have_ } from "../../lib.js";
+import { get } from "../../property";
+import { clamp } from "../../utils";
+
+const basicFruit = $items`orange, grapefruit, grapes, lemon, lime, papaya, cranberries, strawberry, cherry, kumquat, tangerine, raspberry, kiwi, blackberry, banana, cactus fruit, plum, pear, peach`;
```
spaghetti-squash:
```suggestion
const BASIC_FRUIT = $items`orange, grapefruit, grapes, lemon, lime, papaya, cranberries, strawberry, cherry, kumquat, tangerine, raspberry, kiwi, blackberry, banana, cactus fruit, plum, pear, peach`;
```
consts like this typically should be in screaming snake case rather than camelcase. it's a vibes thing

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,144 @@
...
+
+const advFruit = $items`classic banana, antique watermelon, quince`;
+
+function getSeed(classId: number, pathId: number, daycount: number): number {
+  return classId ** 3 + 84 * pathId + 123 * (daycount - 1) + 381;
+}
+
+function getDeterministicDrops(fights: number): number {
+  return fights <= 0 ? 0 : Math.floor((Math.sqrt(8 * fights - 7) - 1) / 2) + 1;
+}
```
spaghetti-squash, line 88:
```suggestion

const fixedDropTurns = Array(11).fill(null).map((_, i) => 1 + (i + 1) * i / 2);

function getDeterministicDrops(fights: number): number {
  const index = fixedDropTurns.findIndex((turn) => turn > fights);
  return index >= 0 ? index : fixedDropTurns.length
}
```

The quadratic formula is a lot less legible than anything other than the quadratic formula.

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    if (isAdv) {
+      pityCount++;
+      pityThreshold = 10;
+      return advFruit[phpMtRand(rng, 0, 2)];
+    }
+
+    pityThreshold += 10;
+    return basicFruit[phpMtRand(rng, 0, 18)];
+  }
```
spaghetti-squash: I was going to comment and say that you shouldn't define this function inside its parent, before I saw the `pityCount` and `pityThreshold`. I still think there's a better way to restructure, but it's not obvious immediately.

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  const deterministicFights = clamp(maxFights, 0, 56);
+  const overage = clamp(maxFights - 56, 0, Infinity);
+
+  for (let i = 0; i < getDeterministicDrops(deterministicFights); i++) {
+    const fight = 1 + (i * (i + 1)) / 2;
+    const rng = phpSeed(getSeed(classId, pathId, daycount) + 381 * (fight - 1));
```
spaghetti-squash: Is creating a new php seed each instance actually Correct here? I thought php rngs persisted stateful information, based on my first reading of the code. Maybe I'm wrong! I do see that fight is included in the seed so I guess that's fine, actually.

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    pityThreshold += 10;
+    return basicFruit[phpMtRand(rng, 0, 18)];
+  }
+
+  const deterministicFights = clamp(maxFights, 0, 56);
+  const overage = clamp(maxFights - 56, 0, Infinity);
+
+  for (let i = 0; i < getDeterministicDrops(deterministicFights); i++) {
+    const fight = 1 + (i * (i + 1)) / 2;
```
spaghetti-squash: You can now recover `fight` from the array in my suggested code for `getDeterministicDrops`

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,144 @@
...
+    const rng = phpSeed(getSeed(classId, pathId, daycount) + 381 * (fight - 1));
+
+    results.set(fight - 1, getFruit(rng));
+  }
+
+  for (let i = 0; i < overage; i++) {
+    const fight = 57 + i;
```
spaghetti-squash: might want to iterate directly over `fight`, rather than over `i` as an interstitial thing.

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,144 @@
...
+  return classId ** 3 + 84 * pathId + 123 * (daycount - 1) + 381;
+}
+
+function getDeterministicDrops(fights: number): number {
+  return fights <= 0 ? 0 : Math.floor((Math.sqrt(8 * fights - 7) - 1) / 2) + 1;
+}
+
+function laughingStockDrops(
+  classId: number,
+  pathId: number,
```
spaghetti-squash: I'd take the class and path args directly and translate them to ids in here, rather than outside.

src/resources/2026/LaughingStock.ts
spaghetti-squash: Maybe a FruitTracker class?

spaghetti-squash (commented):
Is the pity count/threshold the only reason we have to compute all of the drops prior to the point your'e at?

src/resources/2026/LaughingStock.ts
Ignose: I have no idea; phpSeeds are alchemy to me.

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,161 @@
...
+    if (isAdvanced) {
+      this.pityCount++;
+      this.pityThreshold = 10;
+      return ADVANCED_FRUIT[phpMtRand(rng, 0, 2)];
+    }
+
+    this.pityThreshold += 10;
+    return BASIC_FRUIT[phpMtRand(rng, 0, 18)];
+  }
+}
```
spaghetti-squash, line 88: I would probably put the rng into the constructor of this but otherwise cool and woke

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,161 @@
...
+ * @param path The path to use for the RNG seed; defaults to the current path
+ * @param daycount The daycount to use for the RNG seed; defaults to the current daycount
+ * @param maxFights The maximum number of fights to simulate
+ * @returns A Map of fight numbers to the fruit dropped on that fight
+ */
+export function laughingStockDrops(
+  characterClass: Class = myClass(),
+  path: Path = myPath(),
+  daycount: number = myDaycount(),
+  maxFights: number,
```
spaghetti-squash: It's weird to have the only mandatory arg also be the final one. I'm surprised nothing in our workflow has complained about this.

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  for (let i = 0; i < overage; i++) {
+    const fight = 57 + i;
+    const rng = phpSeed(getSeed(classId, pathId, daycount) + 381 * (fight - 1));
+
+    if (phpMtRand(rng, 1, 50) !== 1) continue;
+
+    results.set(fight, fruitTracker.getFruit(rng));
+  }
```
spaghetti-squash:
```suggestion
  const seed = getSeed(classId, pathId, daycount);

  const results = new Map<number, Item>();
  const fruitTracker = new FruitTracker();

  const deterministicFights = clamp(maxFights, 0, 56);
  const overage = clamp(maxFights - 56, 0, Infinity);

  for (const fight of fixedDropTurns.filter(
    (fight) => fight <= deterministicFights,
  )) {
    const rng = phpSeed(seed+ 381 * (fight - 1));

    results.set(fight, fruitTracker.getFruit(rng));
  }

  for (let i = 0; i < overage; i++) {
    const fight = 57 + i;
    const rng = phpSeed(seed + 381 * (fight - 1));

    if (phpMtRand(rng, 1, 50) !== 1) continue;

    results.set(fight, fruitTracker.getFruit(rng));
  }
```

src/resources/2026/LaughingStock.ts
spaghetti-squash: This might not actually work, given that we call it a lot with different rng seeds every time (as fight increments)? idk. Do what feels natural, I guess.

src/resources/2026/LaughingStock.ts
```diff
@@ -0,0 +1,160 @@
...
+ * @param maxFights The maximum number of fights to simulate
+ * @returns A Map of fight numbers to the fruit dropped on that fight
+ */
+export function laughingStockDrops(
+  characterClass: Class = myClass(),
+  path: Path = myPath(),
+  daycount: number = myDaycount(),
+  maxFights: number,
+): Map<number, Item> {
+  const seed = getSeed(classId, pathId, daycount);
```
spaghetti-squash: ahh, my bad:

```suggestion
  const seed = getSeed(characterClass.id, path.id, daycount);
```

Ignose (commented):
This fails test; I think phpSeed cares about order and can't arbitrarily be reloaded? Could be tests.

PR #876, opened 2026-08-18
Add 'walfordBucketItem' to fake item list

libraryaddict:
Kolmafia populates the property with a list of non-items, like "rain" and "ice".
This PR should prevent that property being added as an ItemProperty


libraryaddict:
Also seems to be a problem with "tavernLayout" being ceorced into a number. Even if you explicitly pass `""` to it to declare it as a string.

gausie (approved):
Thanks for the PR!

PR #880, opened 2026-08-20
Add sea modifiers

Ignose:


src/modifierTypes.ts
gausie, line 1: No

gausie:
I can close the PR or you can fix in place here, but this isn't how you do this

PR #885, opened 2026-08-28
use .js import style

spaghetti-squash:


Rinn (approved):
If there's a lint setting to check for this we should enable it

PR #897, opened 2026-09-05
support long else...if blocks?

spaghetti-squash:
This essentially lets us continue to honor the contract of `if_` as it previously existed and as it existed in the first `else` work. But we don't need to honor the latter, because it hasn't been released.

I don't love that the frontmost elements don't need to be an object but subsequent entries do. Maybe it makes sense to keep `if_` as it once was, and do `ifElse` that accepts just an `ElseTrain` argument?

### DIFF
--- a/src/combat.ts
+++ b/src/combat.ts
@@ -578,7 +578,9 @@ export class Macro {
         this.step("else").step(elseEntry);
       }
     }
-    this.step(elseTrain.filter(Boolean).length ? "endelse" : "endif");
+
+    this.step("endif");
+
     return this;
   }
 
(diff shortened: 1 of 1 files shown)


Ignose (approved):
I'm not sure where size limits for macros are housed (maybe Libram lets the user put this particular gun to this foot) but otherwise this looks good.

spaghetti-squash:
> I'm not sure where size limits for macros are housed (maybe Libram lets the user put this particular gun to this foot) but otherwise this looks good.

Libram doesn't track size limits, macros are remarkably stateless. Like, all they contain is an array of bare strings for their steps.

We could adjust the Type of the ElseTrain to be 
```
export type ElseTrain =
  [...Tuple<ElseIfComponent, Range<0, 69>>] | [...Tuple<ElseIfComponent, Range<0, 68>>, string | Macro | undefined]
```

